PR #1, opened 2015-02-26
Initial (Huge) commit

jmgoldb:


seanwolter:
72,208 additions may be a Vokal record

jmgoldb:
Go Big or Go Home

scottferg:
The average @nick-fedesna pull request is at least 85,000

nick-fedesna:
Boilerplate much?  I'm not even going to click `Files changed` as I don't want my computer to explode.

vokal-isaac:
Miraculously, GitHub and Chrome seem willing to actually load the whole diff.  I'll give it at least a quick pass tomorrow and we can talk about any larger things late in the day tomorrow or on Monday.

MH4U Dex/AppDelegate.m
```diff
@@ -0,0 +1,43 @@
...
+
+@implementation AppDelegate
+
+- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
+{
+    NSLog(@"Loading Data");
+    self.coreDataController = [[CoreDataController alloc] init];
+    [self.coreDataController loadMonsterData];
+    [self.coreDataController loadMonsterDropData];
+    [self.coreDataController loadRegionData];
```
vokal-isaac: It might make more sense to use a shared singleton pattern rather than a property on the app delegate.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+#pragma mark - Initial Data Loading
+
+- (void)loadMonsterData
+{
+    // Because monsters are added first, there are no relationships to set up, so we can hydrate them.
+    NSString *path = [[NSBundle mainBundle] pathForResource:@"monsters" ofType:@"json"];
+    NSDictionary *attributes = @{@"id":@"_id", @"monster_class":@"class", @"name":@"name", @"jpn_name":@"jpn_name", @"sort_name":@"sort_name", @"trait":@"trait", @"icon":@"icon_name", @"url":@"url"};
+    [self.managedObjectContext hydrateStoreWithJSONAtPath:path attributeMappings:attributes forEntityName:@"Monster"];
+    path = [[NSBundle mainBundle] pathForResource:@"damageZones" ofType:@"json"];
+    attributes = @{@"id":@"_id", @"monster_id":@"monster_id", @"monsterName":@"monster_name", @"jpnMonsterName":@"jpn_monster_name", @"bodyPart":@"body_part", @"cut":@"cut", @"impact":@"impact", @"shot":@"shot", @"fire":@"fire", @"water":@"water", @"thunder":@"thunder", @"ice":@"ice", @"dragon":@"dragon", @"ko":@"ko", @"extract":@"extract"};
```
vokal-isaac: Some of these lines are way too long; it might also be more readable if it were formatted according to our standards.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+    [self.managedObjectContext hydrateStoreWithJSONAtPath:path attributeMappings:attributes forEntityName:@"DamageZone"];
```
vokal-isaac: With mogenerator, you can call the `entityName` class method on the corresponding class rather than hard-coding the string for the entity name.

Does "Damage Zone" loading conceptually belong inside `loadMonsterData`?

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+
+#pragma mark - Initial Data Loading
+
+- (void)loadMonsterData
+{
+    // Because monsters are added first, there are no relationships to set up, so we can hydrate them.
+    NSString *path = [[NSBundle mainBundle] pathForResource:@"monsters" ofType:@"json"];
+    NSDictionary *attributes = @{@"id":@"_id", @"monster_class":@"class", @"name":@"name", @"jpn_name":@"jpn_name", @"sort_name":@"sort_name", @"trait":@"trait", @"icon":@"icon_name", @"url":@"url"};
+    [self.managedObjectContext hydrateStoreWithJSONAtPath:path attributeMappings:attributes forEntityName:@"Monster"];
+    path = [[NSBundle mainBundle] pathForResource:@"damageZones" ofType:@"json"];
```
vokal-isaac: It's probably worth making a file-level constant for `@"json"`.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+}
+
+- (void)loadMonsterDropData
+{
+    NSArray *dropList = [self loadArrayFromJsonFileNamed:@"monsterDrops"];
+    for (NSDictionary *drop in dropList) {
+        // Only add the drop if it is not already there
+        if (![self checkCountForEntityWithEntityName:@"MonsterDrop" withIDNumber:drop[@"_dropID"]]) {
+            NSEntityDescription *newDropDescription = [NSEntityDescription entityForName:@"MonsterDrop" inManagedObjectContext:self.managedObjectContext];
+            MonsterDrop *newDrop = [[MonsterDrop alloc] initWithEntity:newDropDescription insertIntoManagedObjectContext:self.managedObjectContext];
```
vokal-isaac: mogenerator adds a class method, `insertInManagedObjectContext:` to your managed object subclass that simplifies this a bit.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+            [newDrop setDropBasicValuesWithID:drop[@"_dropID"] method:drop[@"method"] quantity:drop[@"quantity"]];
+            
+            // If the Item for this drop already exists in the persistent store ...
+            Item *item = (Item *)[self getUniqueEntityWithEntityName:@"Item" withNameAttribute:drop[@"item_name"]];
```
vokal-isaac: It'd probably be good to create constants for all these dictionary keys.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+                // Otherwise, we should create the item.
+                NSEntityDescription *newItemDescription = [NSEntityDescription entityForName:@"Item" inManagedObjectContext:self.managedObjectContext];
+                Item *newItem = [[Item alloc] initWithEntity:newItemDescription insertIntoManagedObjectContext:self.managedObjectContext];
+                [newItem.monsterSourceSet addObject:newDrop];
+                newItem.name = drop[@"item_name"];
+                newDrop.item = newItem;
+            }
+            
+            // If this is a monster drop ...
+            if (![drop[@"monster_name"] isEqualToString:@""]) {
```
vokal-isaac: If `drop` or `drop[@"monster_name"]` is `nil`, then `[drop[@"monster_name"] isEqualToString:@""]` will be `nil`, and `![drop[@"monster_name"] isEqualToString:@""]` will be true—is that intended?  Otherwise, you might want something like `drop[@"monster_name"].length`, which will be true whenever `drop[@"monster_name"]` is not `nil` and also has a nonzero length.  You might also want to verify that `drop[@"monster_name"]` is actually an `NSString`, but that may already be guaranteed in the context of this method.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+                // We need to add this drop to the monster's list, if it is not there already.
+                
+                // First we should ensure that the monster already exists in the persistent store
+                Monster *monster = (Monster *)[self getUniqueEntityWithEntityName:@"Monster" withIDNumber:drop[@"monster_id"]];
+                if (!monster) {
+                    NSLog(@"Monster not found! Returning early.");
+                    @throw @"Monster nto found!";
```
vokal-isaac: Throwing an exception is not a mechanism for returning early.  In Objective-C, exceptions should only be used for unrecoverable states where the app should be or already is crashing.  See, e.g., [this Stack Overflow answer](http://stackoverflow.com/a/324805/291280).

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+            return results.firstObject;
+        } else {
+            NSLog(@"Too many entities found!");
+            return nil;
+        }
+    } else {
+        NSLog(@"Error fetching object!");
+        NSLog(@"%@, %@", error, error.localizedDescription);
+        return nil;
+    }
```
vokal-isaac: Given the level of duplication of code across these `getUnique...` methods, it would probably be better to have a more-generic common method that all of these call into to handle the common aspects.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+    if (error) {
+        NSLog(@"Error reading file: %@", error.localizedDescription);
+        return nil;
+    }
+    return (NSArray *)[NSJSONSerialization JSONObjectWithData:[fileContents dataUsingEncoding:NSUTF8StringEncoding]
+                                                      options:0
+                                                        error:nil];
+}
+
+- (BOOL)didSuccessfullySaveContext
```
vokal-isaac: This is misleadingly named—it sounds like a (delegate) method that gets called after the save, but it's actually attempting the save.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+                                                           options:0
+                                                             error:nil];
+}
+
+- (NSArray *)loadArrayFromJsonFileNamed:(NSString *)fileName
+{
+    NSString *filePath = [[NSBundle mainBundle] pathForResource:fileName ofType:@"json"];
+    NSError *error;
+    NSString *fileContents = [NSString stringWithContentsOfFile:filePath encoding:NSUTF8StringEncoding error:&error];
+    if (error) {
```
vokal-isaac: For methods that take an `NSError **` parameter, the value of that parameter is (generally) undefined on success, not guaranteed to be `nil`, so you must check the value returned by the method to determine success.  i.e., `if (!fileContents) {`.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+        NSLog(@"Error reading file: %@", error.localizedDescription);
+        return nil;
+    }
+    return (NSDictionary *)[NSJSONSerialization JSONObjectWithData:[fileContents dataUsingEncoding:NSUTF8StringEncoding]
+                                                           options:0
+                                                             error:nil];
```
vokal-isaac: If you're not passing the address of an `NSError *`, you should be passing `NULL` rather than `nil`, but given that you have error handling already in place in this method, you should actually pass an error object in and check success before returning.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+                    newDrop.monsterSource = monster;
+                    newDrop.rank = drop[@"rank"];
+                    newDrop.percent = drop[@"percent"];
+                }
+            }
+        }
+    }
+    
+    // We should now try to save the manged object context
+    if (![self didSuccessfullySaveContext]) {
```
vokal-isaac: Unless there's a specific reason for saving after each part of the loading process, you might want to wait until everything's loaded before saving—saving can be slow.  That might be a premature optimization at this point, though.

vokal-isaac: You might also not want to attempt the save if the load operation returned no JSON objects.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+    }
+    // If it wasn't already, the area is now attached to the region
+    
+    // We should now consider adding this drop to the area.
+    // First, check if the area drop already exists, by way of checking the idDecimalString
+    NSString *idDecimalString = areaDropDict[@"_id"];
+    AreaDrop *drop = [self getUniqueAreaDropWithDecimalID:idDecimalString];
+    //If the drop already exists...
+    if (drop) {
+        // ... do nothing. (At least for now).
```
vokal-isaac, line 244: If you put in a comment with `TODO:` (the `:` is necessary, I'm not sure if the space after the `:` is necessary), Xcode will usually pick up on it and show it in the drop-down navigator.  It's also convenient to be able to search for a known tag like `TODO` to find all your things to come back to later.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+        drop.idDecimalString = idDecimalString;
+        drop.method = areaDropDict[@"method"];
+        NSString *percent = areaDropDict[@"chance"];
+        drop.percentValue = percent.integerValue;
+        drop.quantity = areaDropDict[@"number"];
+        drop.rank = areaDropDict[@"rank"];
+        
+        // Hook up the area relationship.  We already ensured the area exists.
+        drop.area = area;
+        [area.dropSet addObject:drop];
```
vokal-isaac: You shouldn't need to set this in both directions—setting it in one direction should automatically set the other direction if you've got the inverse relationship set up.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+        return NO;
+    }
+    return YES;
+}
+
+- (AreaDrop *)getUniqueAreaDropWithDecimalID:(NSString *)decimalID
+{
+    NSError *error = nil;
+    NSFetchRequest *request = [[NSFetchRequest alloc] initWithEntityName:@"AreaDrop"];
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"%K == %@", @"idDecimalString", decimalID];
```
vokal-isaac: It's preferable to use the attribute constants defined by mogenerator over hard-coded strings.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+    [request setPredicate:predicate];
+    
+    NSArray *results = [self.managedObjectContext executeFetchRequest:request error:&error];
+    if (!error) {
+        if (results.count == 0) {
+            // Object not found.
+            return nil;
+        } else if (results.count == 1) {
```
vokal-isaac: Since you returned out of the `if` block, you could just go into the next `if` statement without the chaining `else`.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+
+- (NSUInteger)checkCountForEntityWithEntityName:(NSString *)entityName withNameAttribute:(NSString *)name
+{
+    NSError *error = nil;
+    NSFetchRequest *request = [[NSFetchRequest alloc] initWithEntityName:entityName];
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"%K == %@", @"name", name];
+    [request setPredicate:predicate];
+    
+    NSUInteger count = [self.managedObjectContext countForFetchRequest:request error:&error];
+    if (!error) {
```
vokal-isaac: [The docs for `countForFetchRequest:error:`](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/CoreDataFramework/Classes/NSManagedObjectContext_Class/index.html#//apple_ref/occ/instm/NSManagedObjectContext/countForFetchRequest:error:) say that if an error occurs, `count` will be...?

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+            return nil;
+        }
+    } else {
+        NSLog(@"Error fetching object!");
+        NSLog(@"%@, %@", error, error.localizedDescription);
+        return nil;
+    }
+}
+
+- (NSUInteger)checkCountForEntityWithEntityName:(NSString *)entityName withNameAttribute:(NSString *)name
```
vokal-isaac: Since this just returns the count and doesn't apply any testing/checking, I might name this `countForEntity...`.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+    } else {
+        return 0;
+    }
+}
+
+#pragma mark - Boilerplate Core Data Code
+
+@synthesize managedObjectContext = _managedObjectContext;
+@synthesize managedObjectModel = _managedObjectModel;
+@synthesize persistentStoreCoordinator = _persistentStoreCoordinator;
```
vokal-isaac: You probably don't need these `@synthesize` statements.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+    _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:[self managedObjectModel]];
+    NSURL *storeURL = [[self applicationDocumentsDirectory] URLByAppendingPathComponent:@"MH4U_Dex.sqlite"];
+    NSError *error = nil;
+    NSString *failureReason = @"There was an error creating or loading the application's saved data.";
+    if (![_persistentStoreCoordinator addPersistentStoreWithType:NSSQLiteStoreType configuration:nil URL:storeURL options:nil error:&error]) {
+        // Report any error we got.
+        NSMutableDictionary *dict = [NSMutableDictionary dictionary];
+        dict[NSLocalizedDescriptionKey] = @"Failed to initialize the application's saved data";
+        dict[NSLocalizedFailureReasonErrorKey] = failureReason;
+        dict[NSUnderlyingErrorKey] = error;
```
vokal-isaac: Is there a reason to use a mutable dictionary rather than a dictionary literal?

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+    // Create the coordinator and store
+    
+    _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:[self managedObjectModel]];
+    NSURL *storeURL = [[self applicationDocumentsDirectory] URLByAppendingPathComponent:@"MH4U_Dex.sqlite"];
+    NSError *error = nil;
+    NSString *failureReason = @"There was an error creating or loading the application's saved data.";
+    if (![_persistentStoreCoordinator addPersistentStoreWithType:NSSQLiteStoreType configuration:nil URL:storeURL options:nil error:&error]) {
+        // Report any error we got.
+        NSMutableDictionary *dict = [NSMutableDictionary dictionary];
+        dict[NSLocalizedDescriptionKey] = @"Failed to initialize the application's saved data";
```
vokal-isaac: A localized description should probably be localized (`NSLocalizedString()`).

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+        dict[NSLocalizedFailureReasonErrorKey] = failureReason;
+        dict[NSUnderlyingErrorKey] = error;
+        error = [NSError errorWithDomain:@"YOUR_ERROR_DOMAIN" code:9999 userInfo:dict];
```
vokal-isaac, line 400: You probably should replace `@"YOUR_ERROR_DOMAIN"` with a constant that is exported (so that callers can test the error domain) and with a reverse-DNS-style string (e.g., `io.vokal.something`).

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,430 @@
...
+        // Replace this with code to handle the error appropriately.
+        // abort() causes the application to generate a crash log and terminate. You should not use this function in a shipping application, although it may be useful during development.
```
vokal-isaac, line 402: Consider adding a `TODO:` to do what these comments say.

MH4U Dex/DamageZoneCell.h
```diff
@@ -0,0 +1,27 @@
...
+@property (nonatomic, assign) NSInteger cut;
+@property (nonatomic, assign) NSInteger impact;
+@property (nonatomic, assign) NSInteger shot;
+@property (nonatomic, assign) NSInteger ko;
+@property (nonatomic, strong) NSString *extract;
+@property (nonatomic, assign) NSInteger fire;
+@property (nonatomic, assign) NSInteger water;
+@property (nonatomic, assign) NSInteger thunder;
+@property (nonatomic, assign) NSInteger ice;
+@property (nonatomic, assign) NSInteger dragon;
```
vokal-isaac: Is there a model that can be used to contain all these properties?  If so, consider passing the model into the `displayContents` method rather than having a property.

MH4U Dex/DamageZoneCell.m
```diff
@@ -0,0 +1,57 @@
...
+
+@end
+
+@implementation DamageZoneCell
+
+- (void)displayContents
+{
+    self.bodyPartLabel.text = self.bodyPart;
+    self.extractLabel.text = self.extract;
+    if ([self.bodyPart isEqualToString:@"Body Part"]) {
```
vokal-isaac: At a minimum, a special string like this should be a constant, but if this is something that might get localized, you especially don't want to be comparing string values as a test.  Perhaps some kind of `BOOL` flag or something?

MH4U Dex/DamageZoneCell.m
```diff
@@ -0,0 +1,57 @@
...
+        // This is the template.
+        self.cutLabel.text = @"Cut";
+        self.impactLabel.text = @"Impact";
+        self.shotLabel.text = @"Shot";
+        self.koLabel.text = @"KO";
+        self.fireLabel.text = @"Fire";
+        self.iceLabel.text = @"Ice";
+        self.dragonLabel.text = @"Dragon";
+        self.waterLabel.text = @"Water";
+        self.thunderLabel.text = @"Thunder";
```
vokal-isaac, line 43: These should probably all be wrapped with `NSLocalizedString()`.

MH4U Dex/DamageZoneCell.m
```diff
@@ -0,0 +1,57 @@
...
+    } else {
+        self.cutLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.cut];
+        self.impactLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.impact];
+        self.shotLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.shot];
+        self.fireLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.fire];
+        self.waterLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.water];
+        self.thunderLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.thunder];
+        self.iceLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.ice];
+        self.dragonLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.dragon];
+        self.koLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)self.ko];
```
vokal-isaac: Our preferred mechanism for inserting `NSInteger`/`NSUInteger`/etc. into a format string is to use `%@` as the format specifier and `@`-box the value (creating an `NSNumber` object from it).  e.g., 

``` objc
self.koLabel.text = [NSString stringWithFormat:@"%@", @(self.ko)]; 
```

MH4U Dex/DamageZonesCollectionViewController.m
```diff
@@ -0,0 +1,92 @@
...
+
+#import <CoreData/CoreData.h>
+
+#import "DamageZoneCell.h"
+#import "DamageZone.h"
+
+@interface DamageZonesCollectionViewController ()
+
+@property (weak, nonatomic) IBOutlet UICollectionViewFlowLayout *flowLayout;
+@property (nonatomic, strong) NSArray *damageZones;
```
vokal-isaac, line 26: At least at the file level, the order of the qualifiers on properties should be the same.

MH4U Dex/DamageZonesCollectionViewController.m
```diff
@@ -0,0 +1,92 @@
...
+    [fetchRequest setPredicate:predicate];
+    [fetchRequest setSortDescriptors:@[sortDescriptor]];
+    self.damageZones = [self.managedObjectContext executeFetchRequest:fetchRequest error:&fetchError];
+}
+
+#pragma mark <UICollectionViewDataSource>
+
+- (NSInteger)numberOfSectionsInCollectionView:(UICollectionView *)collectionView
+{
+    return 2;
```
vokal-isaac: Use an `enum` to define constants for your sections as well as a section-count constant.

MH4U Dex/DamageZonesCollectionViewController.m
```diff
@@ -0,0 +1,92 @@
...
+}
+
+- (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSection:(NSInteger)section
+{
+    if (section == 0) {
```
vokal-isaac: It's generally better (and more future-proof) to use a `switch`-`case` statement for breaking down what to do in each section.

MH4U Dex/HomeViewController.m
```diff
@@ -0,0 +1,37 @@
...
+
+@interface HomeViewController ()
+
+@end
+
+@implementation HomeViewController
+
+- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
+{
+    if ([segue.identifier isEqualToString:@"showMonsterEncyclopedia"]) {
```
vokal-isaac, line 23: It would probably be good to have constants for these identifiers—perhaps generated from the storyboard using a tool like https://github.com/square/objc-codegenutils#objc-identifierconstants.

MH4U Dex/Info.plist
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>$(EXECUTABLE_NAME)</string>
+	<key>CFBundleIdentifier</key>
+	<string>JMG.$(PRODUCT_NAME:rfc1034identifier)</string>
```
vokal-isaac, line 10: [The docs](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102070) say that `CFBundleIdentifier` should be in reverse-DNS format.

MH4U Dex/ItemAreaSourcesTableViewController.m
```diff
@@ -0,0 +1,90 @@
...
+    if (self.sources.count) {
+        return self.sources.count;
+    } else {
+        return 0;
+    }
+}
+
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    ItemAreaSourceTableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:NSStringFromClass([ItemAreaSourceTableViewCell class])];
```
vokal-isaac: Look at the docs for `dequeueReusableCellWithIdentifier:` and `dequeueReusableCellWithIdentifier:forIndexPath:`.

MH4U Dex/ItemAreaSourcesTableViewController.m
```diff
@@ -0,0 +1,90 @@
...
+    AreaDrop *drop = self.sources[indexPath.row];
+    cell.method = drop.method;
+    Area *area = [drop valueForKey:AreaDropRelationships.area];
```
vokal-isaac: Is there a reason not to use `drop.area`?

MH4U Dex/ItemAreaSourcesTableViewController.m
```diff
@@ -0,0 +1,90 @@
...
+    return cell;
+}
+
+- (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    // ToDo: chage to specific area/region once area/region views are in.
+    UIStoryboard *storyBoard = [UIStoryboard storyboardWithName:@"Main" bundle:nil];
+    RegionEncyclopediaCollectionViewController *regionVC = [storyBoard instantiateViewControllerWithIdentifier:@"RegionEncyclopediaCollectionViewController"];
+    regionVC.managedObjectContext = self.managedObjectContext;
+    [self.navigationController pushViewController:regionVC animated:YES];
```
vokal-isaac, line 86: Could this be handled with a segue?

MH4U Dex/ItemContainerViewController.m
```diff
@@ -0,0 +1,61 @@
...
+    [super viewDidLoad];
+    [self.areaSourceSubView setHidden:YES];
+    self.navigationItem.title = [NSString stringWithFormat:@"%@ Drop Sources", self.itemName];
+    self.itemNameLabel.text = self.itemName;
+}
+
+- (IBAction)segmentedControlChanged:(UISegmentedControl *)segmentedControl
+{
+    switch (segmentedControl.selectedSegmentIndex) {
+        case 0:
```
vokal-isaac: You might want to use an `enum` to declare constants for these indices.

MH4U Dex/ItemContainerViewController.m
```diff
@@ -0,0 +1,61 @@
...
+            [self.monsterSourceSubView setHidden:NO];
```
vokal-isaac: `self.monsterSourceSubView.hidden = NO;`?

MH4U Dex/ItemMonsterSourcesTableViewController.m
```diff
@@ -0,0 +1,92 @@
...
+    NSSortDescriptor *sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:@"id" ascending:YES];
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropRelationships.item, item];
+    [fetchRequest setPredicate:predicate];
+    [fetchRequest setSortDescriptors:@[sortDescriptor]];
+    
+    self.sources = [self.managedObjectContext executeFetchRequest:fetchRequest error:&fetchError];
+    // If the fetch failed (most likely because the item cannot be found from any monsters) ...
+    if (!self.sources) {
+        // Set sources to empty.
+        self.sources = [NSArray arrayWithObjects:nil];
```
vokal-isaac: `@[]` is preferable to `[NSArray arrayWithObjects:nil]`.

MH4U Dex/MonsterEncyclopediaViewController.m
```diff
@@ -0,0 +1,100 @@
...
+                                                                              forIndexPath:indexPath];
+    Monster *monster;
+    switch (indexPath.section) {
+        case 0:
+        monster = (Monster *)self.smallMonsters[indexPath.row];
+        break;
+        case 1:
+        monster = (Monster *)self.largeMonsters[indexPath.row];
+        break;
+    }
```
vokal-isaac, line 101: Indentation within this `switch` doesn't look right.

MH4U Dex/MonsterLowRankTableViewController.h
```diff
@@ -0,0 +1,18 @@
...
+//
+
+#import <UIKit/UIKit.h>
+
+@interface MonsterLowRankTableViewController : UITableViewController
+
+@property (nonatomic, strong) NSManagedObjectContext *managedObjectContext;
+@property (nonatomic, strong) NSString *monster;
+
+- (void)setMonster:(NSString *)monster;
```
vokal-isaac: This is already implied by `@property (nonatomic, strong) NSString *monster;`.

MH4U Dex/MonsterOverviewViewController.m
```diff
@@ -0,0 +1,43 @@
...
+
+@end
+
+@implementation MonsterOverviewViewController
+
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    self.monsterNameLabel.text = self.monster;
+    NSString *imageName = [NSString stringWithFormat:@"%@-icon", [self.monster lowercaseString]];
```
vokal-isaac, line 27: If the image name for a monster is always going to be structured this way, consider moving at least the image name, if not the whole image, into a readonly property on the model object (on your subclass, not in Core Data).

vokal-isaac:
Overall:
- If I commented on something, you probably should see if there's similar or identical code elsewhere to which the comment would also apply.
- If you're writing the same or nearly the same code more than once, you should at least consider whether that code can be generalized into a method/base class/something somewhere so that you don't repeat yourself (DRY).
- All user-facing strings should be wrapped in `NSLocalizedString()`.  Consider using one or more classes and class methods to centralize all your user-facing strings and calls to `NSLocalizedString()`, as in, e.g., https://github.com/vokal/ThousandTrails-iOS/blob/master/ThousandTrails/ThousandTrails/Shared%20Utility/TTRString.h
- Avoid "stringly-typed" code whenever possible.
- Ordering your `#import` statements from the start is easier than going back and re-ordering them later.
- Organize your files into subdirectories and corresponding groups in Xcode.  Take a look at various client projects for some examples of potential directory structures.

Let's plan to touch base on all this Monday afternoon—put a half-hour meeting on our calendars at a time when we're both free for sometime that afternoon.

MH4U Dex/CoreDataController.m
jmgoldb: Each monster has its own set of damage zones, so I think it makes sense to load damage zones where they currently are.

MH4U Dex/AppDelegate.m
jmgoldb: Changed.  This singleton pattern is really useful, now I don't have to worry about passing the managedObjectContext everywhere!

MH4U Dex/CoreDataController.m
jmgoldb: Changed

MH4U Dex/CoreDataController.m
jmgoldb: Changed.  This is way more convenient.

MH4U Dex/CoreDataController.m
jmgoldb: Changed

MH4U Dex/CoreDataController.m
jmgoldb: Changed here, and throughout the project where appropriate

MH4U Dex/CoreDataController.m
jmgoldb: drop[@"monster_name"] is guaranteed to be a string, but evaluating length is a good solution to if drop is nil.  Changed!

MH4U Dex/CoreDataController.m
jmgoldb: Removed the throw, but It might not be a bad idea to revisit this (and other error handling) down the line.

MH4U Dex/CoreDataController.m
jmgoldb: Made a generic method, not sure why the diff isn't marked as outdated

MH4U Dex/CoreDataController.m
jmgoldb: Renamed

MH4U Dex/CoreDataController.m
jmgoldb: That makes sense!  Changed

MH4U Dex/CoreDataController.m
jmgoldb: I'll wait to do this optimization later. Right now the only delay is caused by the program failing to find certain mock JSON files that I haven't yet created.

MH4U Dex/CoreDataController.m
jmgoldb: I've fixed this where I've noticed it, but it's possible I've missed a few spots.

MH4U Dex/CoreDataController.m
jmgoldb: The boilerplate error handling should probably be revisited later anyways, so I'm not going to worry about localizing the error description yet

MH4U Dex/CoreDataController.m
jmgoldb, line 400: Same as above.

MH4U Dex/DamageZoneCell.h
jmgoldb: Changed.  I still need to do this with the other cells, but the other cells need a bit of a revamp anyways.  Moving forward, I'll be sure to pass the model into displayContents to avoid burdening the cell itself with a bunch of properties.

MH4U Dex/DamageZoneCell.m
jmgoldb: I'd like to later gut this out and replace it with a static header, so I'm leaving this as-is for now.

MH4U Dex/DamageZoneCell.m
jmgoldb, line 43: Same as above.

MH4U Dex/DamageZonesCollectionViewController.m
jmgoldb, line 26: I always prefer having atomic/nonatomic before strong/weak, but generating IBOutlets via interface builder always sets them up the other way.  I forgot to swap this one around.

MH4U Dex/DamageZonesCollectionViewController.m
jmgoldb: Changed.  The thousand trails code made a great reference for this

MH4U Dex/HomeViewController.m
jmgoldb, line 23: I think this would be nice, but I wasn't really sure on how to install this tool.  Maybe we can go over it later?  I think it would be nice to set aside some time to have all identifiers be storyboard-generated constants.

MH4U Dex/Info.plist
jmgoldb, line 10: Since this technically belongs to Vokal, I'm not even sure what I should do with this.  It was auto-generated.

MH4U Dex/ItemAreaSourcesTableViewController.m
jmgoldb: Ooops!  Fixed.

MH4U Dex/ItemAreaSourcesTableViewController.m
jmgoldb: Changed.  I think I made this same mistake elsewhere, I'll have to look and see where else I made it.

MH4U Dex/ItemAreaSourcesTableViewController.m
jmgoldb, line 86: If I handle this via the storyboard, it creates a bunch of odd behavior, because I have a cycle in my graph of views that it doesn't seem to like.  Doing things this way makes things way cleaner.

MH4U Dex/ItemContainerViewController.m
jmgoldb: _Annoyed grunt_

MH4U Dex/ItemMonsterSourcesTableViewController.m
jmgoldb: Changed.  What about [NSArray array]?  (Or, for a mutable array, [NSMutableArray array]?)

MH4U Dex/MonsterEncyclopediaViewController.m
jmgoldb, line 101: Weird, but fixed.

MH4U Dex/MonsterOverviewViewController.m
jmgoldb, line 27: I'm a little confused by this comment, so I'll revisit this later, once I start adding in more images.

MH4U Dex/Info.plist
vokal-isaac, line 10: `io.vokal.$(PRODUCT_NAME:rfc1034identifier)` would probably work.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,446 @@
...
+
+@property (readwrite, strong, nonatomic) NSManagedObjectContext *managedObjectContext;
+@property (readwrite, strong, nonatomic) NSManagedObjectModel *managedObjectModel;
+@property (readwrite, strong, nonatomic) NSPersistentStoreCoordinator *persistentStoreCoordinator;
+
+@end
+
+@implementation CoreDataController
+
+NSString *const JSON = @"json";
```
vokal-isaac, line 29: Constants that are only used within the `.m` file should be declared `static`; if the constant is used outside the `.m` file, it needs to be exported (`FOUNDATION_EXPORT`) in the corresponding header and the name needs to be prefixed with at least the project's three-letter prefix to avoid naming collisions.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,446 @@
...
+
+#pragma mark - Item Constants
```
vokal-isaac, line 31: These `#pragma mark` statements might not be useful—it's worth checking the symbol navigator within the file to see how they display there and whether any of the constants you're declaring between `#pragma mark` statements show up there.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,446 @@
...
+    NSString *path = [[NSBundle mainBundle] pathForResource:MonstersFileName ofType:JSON];
+    NSDictionary *attributes = @{
+                                 @"id":@"_id",
+                                 @"monster_class":@"class",
+                                 @"name":@"name",
+                                 @"jpn_name":@"jpn_name",
+                                 @"sort_name":@"sort_name",
+                                 @"trait":@"trait",
+                                 @"icon":@"icon_name",
+                                 @"url":@"url"
```
vokal-isaac, line 103: Space after each `:`; trailing comma on the last item.

vokal-isaac, line 103: Also, aren't there now constants for some or all of these strings?

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,446 @@
...
+    [self.managedObjectContext hydrateStoreWithJSONAtPath:path attributeMappings:attributes forEntityName:[DamageZone entityName]];
+}
+
+- (void)loadMonsterDropData
+{
+    NSArray *dropList = [self loadArrayFromJsonFileNamed:MonsterDropsFileName];
+    for (NSDictionary *drop in dropList) {
+        // Only add the drop if it is not already there
+        NSNumber *monsterDropID = drop[MonsterDropIDKey];
+        NSPredicate *monsterDropPredicate = [NSPredicate predicateWithFormat:@"%K == %d", MonsterDropIDKey, monsterDropID.intValue];
```
vokal-isaac, line 133: For things that are properties of your core data model entity, you should have a constant provided by mogenerator—in this case, something like `MonsterAttributes.id`.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,446 @@
...
+            
+            // If the Item for this drop already exists in the persistent store ...
+            NSPredicate *itemPredicate = [NSPredicate predicateWithFormat:@"%K == %@", ItemNameKey, drop[MonsterDropItemNameKey]];
+            Item *item = (Item *)[self getUniqueEntityWithEntityName:[Item entityName] withPredicate:itemPredicate];
+            if (item) {
+                // We need to hook up relationships appropriately.
+                newDrop.item = item;
+            } else {
+                // Otherwise, we should create the item.
+                Item *newItem = [Item insertInManagedObjectContext:self.managedObjectContext];
```
vokal-isaac, line 148: If you used the `item` variable that already exists, you could move the assignment that's basically duplicated as the last line of both the `if` and `else` blocks outside the conditional.

MH4U Dex/CoreDataController.m
```diff
@@ -0,0 +1,446 @@
...
+        // Only add the region if it is not already there
+        if (!region) {
+            region = [Region insertInManagedObjectContext:self.managedObjectContext];
+            region.name = regionDict[RegionNameKey];
+            region.id = regionDict[RegionIDJSONKey];
+            region.keyName = regionDict[RegionKeyNameKey];
+        }
+        // regardless if the region is new or not, we should populate it with area data
+        // But first, we need to check that the region file exists
+        NSArray *areaDrops = [self loadArrayFromJsonFileNamed:[NSString stringWithFormat:@"%@%@", region.keyName, RegionDropsFileNameSuffix]];
```
vokal-isaac, line 198: `[region.keyName stringByAppendingString:RegionDropsFileNameSuffix]`

README.md
```diff
@@ -1,2 +1,28 @@
 # MH4U-Dex
 Joseph Goldberg's apprentice project. 
+
+Open-Source under MIT License: http://opensource.org/licenses/MIT
```
vokal-isaac, line 4: Consider linking to the `LICENSE` file in addition to or instead of the generic license at this URL.

README.md
```diff
@@ -1,2 +1,28 @@
...
+
+App is still currently in-development, and undergoing a major refactor.  Current functionality: App reads in monsters and their hitzones/damage resistances, and displays these.  App also is capable of reading in monster drops, and gather locations of items, and listing monster and gather sources for items.  App currently is capable of processing detailed data of these "drops", but currently displays only the very basics.  Navigation to the source monster is possible for monster drops, but navigation to the source area for a gather point is not yet implemented.  (It is next on the list, after this refactor.)
+
+Currently, the app makes no distinction when displaying data, between low, high, and g-rank, but this should not be difficult to implement, as this information IS tracked on the model-side of things.
```
vokal-isaac, line 8: Look into markdown formatting to make the rendered version of this more readable.

MH4U Dex/MonsterDetailsViewController.m
```diff
@@ -0,0 +1,72 @@
...
+//  Copyright (c) 2015 Joseph Goldberg. All rights reserved.
+//
+
+#import "MonsterDetailsViewController.h"
+#import "MonsterOverviewViewController.h"
+#import "MonsterLowRankTableViewController.h"
+
+typedef NS_ENUM(NSInteger, MonsterSegmentedControlPage) {
+    Overview = 0,
+    LowRank = 1,
```
vokal-isaac, line 15: These names might be a bit too generic—they don't really indicate that the constants refer to the monster segmented control indicies.

MH4U Dex/CoreDataController.m
vokal-isaac: Error handling is definitely something that should be done; `@throw` isn't a mechanism for handling (non-crashing) errors, though.

MH4U Dex/CoreDataController.m
vokal-isaac: Add a `TODO` to revisit, then.

MH4U Dex/DamageZoneCell.m
vokal-isaac: Add a `TODO`.

MH4U Dex/HomeViewController.m
vokal-isaac, line 23: I haven't used that one in particular, but we can look at it.

MH4U Dex/ItemAreaSourcesTableViewController.m
vokal-isaac, line 86: In that case, you may be creating the potential for a very deep navigation stack, which may or may not be an issue...

MH4U Dex/ItemContainerViewController.m
vokal-isaac: [The fourth bullet here](https://github.com/vokal/Coding-Standards/tree/master/iOS#implementation-styling) says, in part:

> Use dot notation for properties

[The sixth bullet here](https://github.com/vokal/Coding-Standards/tree/master/iOS#method-styling) says:

> Use dot syntax for accessing properties and only for accessing properties

MH4U Dex/ItemMonsterSourcesTableViewController.m
vokal-isaac: `[NSArray array]` is equivalent to `@[]` but `@[]` is shorter; if you need an empty mutable array, `[NSMutableArray array]` is a good choice, although if you have a reasonable guess as to the number of things you're going to put into the mutable array, `[NSMutableArray arrayWithCapacity:]` is a better choice.

vokal-isaac:
@jmgoldb  Let's merge this down as-is right now and you can address any remaining comments with changes in your next PR.

@seanwolter @designatednerd @chillpop merge please?

PR #2, opened 2015-03-12
Fundamentals of regions are in.  Also, a lot of new JSON files.

jmgoldb:
Making a pull request early so we can get these JSON files.  Next I am going to set up the distinction of low, high, and g-rank region views, by subclassing the view controller I made.  I considered instead just having one view controller able to handle everything, but that would lead to needless complexity.


MH4U Dex/AreaDropTableViewCell.h
```diff
@@ -0,0 +1,17 @@
...
+//  AreaDropTableViewCell.h
+//  MH4U Dex
+//
+//  Created by Joseph Goldberg on 3/11/15.
+//  Copyright (c) 2015 Joseph Goldberg. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+#import "AreaDrop.h"
```
vokal-isaac: It's usually preferable to forward-declare the class in the header file (`@class AreaDrop;`), rather than importing that class's header, unless there's something more than the class symbol that's necessary for this header.

designatednerd: (you do still have to `#import` in the `.m` file though)

MH4U Dex/CoreDataController.h
```diff
@@ -18,6 +18,8 @@
 - (void)loadRegionData;
 - (BOOL)attemptSaveContext;
 
+- (NSManagedObject *)getUniqueEntityWithEntityName:(NSString *)entityName withPredicate:(NSPredicate *)predicate;
```
vokal-isaac: This (and, actually, most if not all methods exposed in headers) should have a documenting comment.

MH4U Dex/CoreDataController.m
```diff
@@ -130,7 +130,7 @@ - (void)loadMonsterDropData
     for (NSDictionary *drop in dropList) {
         // Only add the drop if it is not already there
         NSNumber *monsterDropID = drop[MonsterDropIDKey];
-        NSPredicate *monsterDropPredicate = [NSPredicate predicateWithFormat:@"%K == %d", MonsterDropIDKey, monsterDropID.intValue];
+        NSPredicate *monsterDropPredicate = [NSPredicate predicateWithFormat:@"%K == %d", MonsterAttributes.id, monsterDropID.intValue];
```
vokal-isaac, line 170: With this (and the similar changes below), are there any constants at the top of this file that aren't used at all anymore?

MH4U Dex/Images.xcassets/Region Icons/ancestral_steppe-icon.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "scale" : "1x",
+      "filename" : "Image.png"
```
vokal-isaac: Did this image come from somewhere?  Are there any potential licensing issues with including it?

MH4U Dex/ItemAreaSourcesTableViewController.m
```diff
@@ -79,10 +79,12 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
 
 - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
 {
-    // ToDo: chage to specific area/region once area/region views are in.
     UIStoryboard *storyBoard = [UIStoryboard storyboardWithName:@"Main" bundle:nil];
-    RegionEncyclopediaCollectionViewController *regionVC = [storyBoard instantiateViewControllerWithIdentifier:@"RegionEncyclopediaCollectionViewController"];
-    regionVC.managedObjectContext = self.managedObjectContext;
+    AreaDrop *drop = self.sources[indexPath.row];
```
vokal-isaac: With table view controllers like this, where you're accessing a property array, I often find it useful to declare helper method(s) to get the model object for a given index path—e.g., `- (AreaDrop *)dropAtIndexPath:(NSIndexPath *)indexPath`—to centralize that logic and to give you a correctly-typed result from the method call (rather than by assigning the item into a typed variable).

MH4U Dex/RegionContainerViewController.m
```diff
@@ -0,0 +1,46 @@
...
+
+#import "RegionContainerViewController.h"
+
+#import "RegionItemTableViewController.h"
+
+#import <CoreData/CoreData.h>
+
+#import "CoreDataController.h"
+
+//NSString *const RegionNameKey = @"name";
```
vokal-isaac: Does this still need to be here?

MH4U Dex/RegionContainerViewController.m
```diff
@@ -0,0 +1,46 @@
...
+@property (nonatomic, weak) IBOutlet UILabel *regionNameLabel;
+
+@end
+
+@implementation RegionContainerViewController
+
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    //[self.regionMapSubview setHidden:YES];
```
vokal-isaac, line 43: Is this still needed?

MH4U Dex/RegionEncyclopediaCollectionViewController.m
```diff
@@ -56,10 +59,27 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
...
+- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
+{
+    if ([segue.identifier isEqualToString:@"showRegionDetails"]) {
+        RegionContainerViewController *regionVC = (RegionContainerViewController *)segue.destinationViewController;
+        regionVC.managedObjectContext = self.managedObjectContext;
+        regionVC.regionName = [sender regionName];
+        regionVC.regionKeyName = [sender regionKeyName];
+        
+        CoreDataController *coreDataController = [CoreDataController sharedCDController];
+        NSPredicate *predicate = [NSPredicate predicateWithFormat:@"%K == %@", @"name", [sender regionName]];
```
vokal-isaac: Does mogenerator provide you a constant you can use in place of the `@"name"` string literal?

MH4U Dex/RegionEncylcopediaCollectionViewCell.m
```diff
@@ -8,6 +8,26 @@
...
 
+- (void)displayContents
+{
+    NSString *imageName = [NSString stringWithFormat:@"%@-icon", self.regionKeyName];
+    UIImage *image = [UIImage imageNamed:imageName];
+    if (image) {
+        self.regionIconImageView.image = image;
+    } else {
+        self.regionIconImageView.image = nil;
+    }
```
vokal-isaac: The `else` clause here is basically "If `image` is `nil`, set `self.regionIconImageView.image` to `nil`", which could just as easily be setting `self.regionIconImageView.image` to `image`—so you don't really need the `if`-`else`, unless there's something else coming here?

MH4U Dex/RegionItemTableViewController.m
```diff
@@ -0,0 +1,87 @@
...
+    NSFetchRequest *areaFetchRequest = [[NSFetchRequest alloc] initWithEntityName:[Area entityName]];
+    NSPredicate *areaPredicate = [NSPredicate predicateWithFormat:@"%K == %@", @"region", self.region];
+    //TODO: Support all ranks.
+    NSPredicate *rankPredicate = [NSPredicate predicateWithFormat:@"%K == %@", @"rank", @"Low"];
+    NSCompoundPredicate *compoundPredicate = [NSCompoundPredicate andPredicateWithSubpredicates:@[rankPredicate, areaPredicate]];
+    [areaFetchRequest setPredicate:compoundPredicate];
+    [areaFetchRequest setSortDescriptors:@[areaSortDescriptor]];
+    self.areaArray = [coreDataController.managedObjectContext executeFetchRequest:areaFetchRequest error:&fetchError];
+    
+
```
vokal-isaac: Extra blank line(s).

MH4U Dex/RegionItemTableViewController.m
```diff
@@ -0,0 +1,87 @@
...
+@property (nonatomic, strong) NSMutableDictionary *areaDrops;
+
+@end
+
+@implementation RegionItemTableViewController
+
+- (void)viewDidLoad
+{
+    CoreDataController *coreDataController = [CoreDataController sharedCDController];
+    self.areaDrops = [NSMutableDictionary dictionary];
```
vokal-isaac: It doesn't look like `self.areaDrops` gets used anywhere.

MH4U Dex/AreaDropTableViewCell.h
jmgoldb: So would that look something like:

#import "AreaDrop.m"

@class AreaDrop;

MH4U Dex/RegionItemTableViewController.m
```diff
@@ -0,0 +1,87 @@
...
+    NSError *fetchError = nil;
+    NSSortDescriptor *areaSortDescriptor = [NSSortDescriptor sortDescriptorWithKey:@"name" ascending:YES];
+    NSFetchRequest *areaFetchRequest = [[NSFetchRequest alloc] initWithEntityName:[Area entityName]];
+    NSPredicate *areaPredicate = [NSPredicate predicateWithFormat:@"%K == %@", @"region", self.region];
+    //TODO: Support all ranks.
+    NSPredicate *rankPredicate = [NSPredicate predicateWithFormat:@"%K == %@", @"rank", @"Low"];
```
vokal-isaac: Could you use mogenerator-provided constants instead of the string literal here and two lines up and two more lines up?

MH4U Dex/Images.xcassets/Region Icons/ancestral_steppe-icon.imageset/Contents.json
jmgoldb: Thanks for this catch!  It came from a wiki with CC-BY-SA 3.0.  It's fine to include it, but I'll have to include an attributions page somewhere.

I think I should go through every image I have, and every Image I plan to add, and check the license on all of them.

MH4U Dex/RegionItemTableViewController.m
```diff
@@ -0,0 +1,87 @@
...
+    Area *area = self.areaArray[section];
+    NSSet *drops = area.drop;
+    return drops.count;
+}
+
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    Area *area = self.areaArray[indexPath.section];
+    NSSortDescriptor *dropSortDescriptor = [NSSortDescriptor sortDescriptorWithKey:@"idDecimalString" ascending:YES];
+    NSArray *dropsArray = [area.drop sortedArrayUsingDescriptors:@[dropSortDescriptor]];
```
vokal-isaac: This may lead to performance issues.

MH4U Dex/RegionContainerViewController.m
jmgoldb, line 43: I actually will later uncomment this, so this one I'll leave as-is.  The earlier one can go, though.

MH4U Dex/RegionEncyclopediaCollectionViewController.m
jmgoldb: That would be RegionAttributes.name, yes?

MH4U Dex/RegionItemTableViewController.m
```diff
@@ -0,0 +1,87 @@
...
+}
+
+- (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    UIStoryboard *storyBoard = [UIStoryboard storyboardWithName:@"Main" bundle:nil];
+    ItemContainerViewController *itemVC = [storyBoard instantiateViewControllerWithIdentifier:@"ItemContainerViewController"];
+    CoreDataController *coreDataController = [CoreDataController sharedCDController];
+    itemVC.managedObjectContext = coreDataController.managedObjectContext;
+    Area *area = self.areaArray[indexPath.section];
+    NSSortDescriptor *dropSortDescriptor = [NSSortDescriptor sortDescriptorWithKey:@"idDecimalString" ascending:YES];
```
vokal-isaac: Is there a constant you could use in place of this string literal?

MH4U Dex/RegionEncylcopediaCollectionViewCell.m
jmgoldb: That's true, but for me at least, writing it this way makes it seem clear to me that leaving the image blank is unambiguously what should happen if the image is not found.

MH4U Dex/AreaDropTableViewCell.h
vokal-isaac: Depending on how it's used in the implementation file, it may or may not be necessary to add `#import "AreaDrop.h"` in `AreaDropTableViewCell.m`—Xcode generally gives an error talking about forward-declarations if you try to do something with the class that requires the import.

MH4U Dex/RegionItemTableViewController.m
jmgoldb: You're right, I should find some way to reorganize things so this is all set up in advance in viewDidLoad, or something.

MH4U Dex/Images.xcassets/Region Icons/ancestral_steppe-icon.imageset/Contents.json
vokal-isaac: CC-BY-SA 3.0 includes this (SA) provision:

> ShareAlike — If you remix, transform, or build upon the material, you must distribute your contributions under the same license as the original.

which I would take to mean that you'd have to distribute this whole app as CC-BY-SA 3.0.

MH4U Dex/RegionContainerViewController.m
vokal-isaac, line 43: Then add at least a comment-note to that effect and perhaps a `TODO:`, depending on why it's commented out and when/why it'll be re-added.

MH4U Dex/Images.xcassets/Region Icons/ancestral_steppe-icon.imageset/Contents.json
jmgoldb: Oh!   Hrm...  

That could be a problem, I'll have to rethink that, or just leave the maps out in a worst-case scenario.

MH4U Dex/RegionContainerViewController.m
jmgoldb, line 43: Will do.

MH4U Dex/RegionEncyclopediaCollectionViewController.m
vokal-isaac: That seems likely from context, but you should be able to verify that by checking its value and verifying that it works.

MH4U Dex/Images.xcassets/Region Icons/ancestral_steppe-icon.imageset/Contents.json
vokal-isaac: Given that, it might be better to take the maps out for now.

jmgoldb: A big issue will be the monster icons, though.  I don't think I can get away with leaving those out.  I'll have to figure something out.

jmgoldb:
I think I've addressed all the remaining pull request issues.

MH4U Dex/RegionItemTableViewController.m
```diff
@@ -0,0 +1,107 @@
...
+    } else {
+        rankPredicate = [NSPredicate predicateWithFormat:@"%K == %@", AreaAttributes.rank, @"Low"];
+    }
+    NSCompoundPredicate *compoundPredicate = [NSCompoundPredicate andPredicateWithSubpredicates:@[rankPredicate, areaPredicate]];
+    [areaFetchRequest setPredicate:compoundPredicate];
+    [areaFetchRequest setSortDescriptors:@[areaSortDescriptor]];
+    self.areaArray = [coreDataController.managedObjectContext executeFetchRequest:areaFetchRequest error:&fetchError];
+    self.areaDropsDict = [NSMutableDictionary dictionary];
+    for (Area *area in self.areaArray) {
+        NSSortDescriptor *dropSortDescriptor = [NSSortDescriptor sortDescriptorWithKey:AreaDropAttributes.idDecimalString ascending:YES];
```
vokal-isaac: Since this doesn't depend on the specific value of `area`, it would be more efficient to move it outside the loop.

MH4U Dex/RegionItemTableViewController.m
```diff
@@ -0,0 +1,107 @@
...
+        NSArray *dropsArray = [area.drop sortedArrayUsingDescriptors:@[dropSortDescriptor]];
```
vokal-isaac: Actually, you might create the array `@[dropSortDescriptor]` outside the loop, too.

MH4U Dex/MonsterDetailsViewController.m
```diff
@@ -29,39 +34,61 @@ @implementation MonsterDetailsViewController
...
+        case HighRank:
+            self.overviewSubView.hidden = YES;
+            self.lowRankSubView.hidden = YES;
+            self.highRankSubView.hidden = NO;
+            self.GRankSubView.hidden = YES;
+        case GRank:
+            self.overviewSubView.hidden = YES;
+            self.lowRankSubView.hidden = YES;
+            self.highRankSubView.hidden = YES;
+            self.GRankSubView.hidden = NO;
```
vokal-isaac, line 63: Missing `break` for this case.

MH4U Dex/MonsterDetailsViewController.m
```diff
@@ -29,39 +34,61 @@ @implementation MonsterDetailsViewController
...
+            self.lowRankSubView.hidden = NO;
+            self.highRankSubView.hidden = YES;
+            self.GRankSubView.hidden = YES;
             self.navigationItem.title = [NSString stringWithFormat:@"%@ LR Drops", self.monster];
             break;
+        case HighRank:
+            self.overviewSubView.hidden = YES;
+            self.lowRankSubView.hidden = YES;
+            self.highRankSubView.hidden = NO;
+            self.GRankSubView.hidden = YES;
```
vokal-isaac: Missing `break` for this case.

MH4U Dex/MonsterDetailsViewController.m
```diff
@@ -29,39 +34,61 @@ @implementation MonsterDetailsViewController
...
+        case GRank:
+            self.overviewSubView.hidden = YES;
+            self.lowRankSubView.hidden = YES;
+            self.highRankSubView.hidden = YES;
+            self.GRankSubView.hidden = NO;
         default:
             break;
     }
```
vokal-isaac, line 68: It might be more concise to set all four `.hidden` properties to `YES` above the `switch`-`case` statement, so that you only have to set one `.hidden` property to `NO` in each case.

MH4U Dex/MonsterDetailsViewController.m
```diff
@@ -29,39 +34,61 @@ @implementation MonsterDetailsViewController
...
             self.navigationItem.title = [NSString stringWithFormat:@"%@ Overview", self.monster];
             break;
         case LowRank:
-            [self.overviewSubView setHidden:YES];
-            [self.lowRankSubView setHidden:NO];
+            self.overviewSubView.hidden = YES;
+            self.lowRankSubView.hidden = NO;
+            self.highRankSubView.hidden = YES;
+            self.GRankSubView.hidden = YES;
             self.navigationItem.title = [NSString stringWithFormat:@"%@ LR Drops", self.monster];
```
vokal-isaac: Check into the difference between setting `self.navigationItem.title` and `self.title` to make sure that you're setting the correct one.

MH4U Dex/MonsterItemTableViewController.m
```diff
@@ -1,45 +1,56 @@
...
-    [fetchRequest setPredicate:predicate];
+    NSSortDescriptor *sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:MonsterDropAttributes.method ascending:YES];
+    NSPredicate *dropPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropRelationships.monsterSource, monster];
+    NSPredicate *rankPredicate;
+    //TODO: Clean this up a little bit.
+    if (self.rank) {
+        rankPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropAttributes.rank, self.rank];
+    } else {
+        rankPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropAttributes.rank, @"Low"];
+    }
```
vokal-isaac: You could rewrite these 5 lines as:

``` objc
rankPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropAttributes.rank, self.rank ?: @"Low"];
```

The `?:` operator returns the first input if it's not `nil`/false-y, otherwise it returns the second input.

Is there a constant for `@"Low"`?  If not, would it make sense to define one?

MH4U Dex/MonsterDetailsViewController.m
```diff
@@ -29,39 +34,61 @@ @implementation MonsterDetailsViewController
 - (void)viewDidLoad
 {
     [super viewDidLoad];
-    [self.lowRankSubView setHidden:YES];
+    self.overviewSubView.hidden = NO;
+    self.lowRankSubView.hidden = YES;
+    self.highRankSubView.hidden = YES;
+    self.GRankSubView.hidden = YES;
```
vokal-isaac: Could you call `[self segmentedControlChanged:]` here rather than duplicating what it does for one case?

MH4U Dex/MonsterDetailsViewController.m
```diff
@@ -29,39 +34,61 @@ @implementation MonsterDetailsViewController
...
+        lowRankVC.rank = @"Low";
+    } else if ([segue.identifier isEqualToString:@"embedHighRank"]) {
+        MonsterItemTableViewController *highRankVC = (MonsterItemTableViewController *)segue.destinationViewController;
+        highRankVC.managedObjectContext = self.managedObjectContext;
+        highRankVC.monster = self.monster;
+        highRankVC.rank = @"High";
+    } else if ([segue.identifier isEqualToString:@"embedGRank"]) {
+        MonsterItemTableViewController *GRankVC = (MonsterItemTableViewController *)segue.destinationViewController;
+        GRankVC.managedObjectContext = self.managedObjectContext;
+        GRankVC.monster = self.monster;
```
vokal-isaac: These three lines are basically identical, except for the variable name, in each block of this `if`-`else`-`if` chain—you could probably move some form of it out to before the `if`s or otherwise restructure the `if`s to avoid the repetition.

MH4U Dex/MonsterItemTableViewController.m
```diff
@@ -1,45 +1,56 @@
...
+    NSCompoundPredicate *compoundPredicate = [NSCompoundPredicate andPredicateWithSubpredicates:@[dropPredicate, rankPredicate]];
+    [fetchRequest setPredicate:compoundPredicate];
     [fetchRequest setSortDescriptors:@[sortDescriptor]];
```
vokal-isaac: `predicate` and `sortDescriptors` are properties, so you should use dot notation.

MH4U Dex/CoreDataController.m
```diff
@@ -26,53 +30,87 @@ @interface CoreDataController ()
 
 @implementation CoreDataController
 
-NSString *const JSON = @"json";
+static NSString *const JSON = @"json";
+static NSString *const FalseString = @"FALSE";
```
vokal-isaac: Is this used anywhere?

MH4U Dex/CoreDataController.m
```diff
@@ -177,6 +210,50 @@ - (void)loadMonsterDropData
...
 }
 
+- (void)loadQuestData
+{
+    NSArray *questList = [self loadArrayFromJsonFileNamed:QuestsFileName];
+    
+    for (NSDictionary *questDict in questList) {
+        Quest *quest;
+        NSNumber *questID = questDict[QuestIDKey];
+        NSPredicate *questPredicate = [NSPredicate predicateWithFormat:@"%K == %d", QuestAttributes.id, questID.intValue];
```
vokal-isaac: It would be better to use `questID` and `%@`—using `.intValue` and `%d` limits the range of values that will work more than using the `NSNumber` object directly.

MH4U Dex/CoreDataController.m
```diff
@@ -363,6 +439,87 @@ - (NSUInteger)countForEntityWithEntityName:(NSString *)entityName withPredicate:
...
+    drop.item = item;
+    return drop;
+}
+
+- (Quest *)questFromDictionary:(NSDictionary *)questDict
+{
+    Quest *quest = [Quest insertInManagedObjectContext:self.managedObjectContext];
+    quest.id = questDict[QuestIDKey];
+    quest.name = questDict[QuestNameKey];
+    if ([questDict[QuestCaravanHallKey] isEqualToString:QuestCaravanKey]) {
```
vokal-isaac, line 471: Why are these coming through as `NSString` objects in the JSON rather than `NSNumber` objects with boolean values?

MH4U Dex/CoreDataController.m
```diff
@@ -363,6 +439,87 @@ - (NSUInteger)countForEntityWithEntityName:(NSString *)entityName withPredicate:
...
+    quest.firstMonster = (Monster *)[self getUniqueEntityWithEntityName:[Monster entityName] withPredicate:monsterPredicate];
+    if (quest.firstMonster) {
+        //Continue on to the second, and so on.  By nesting this way, we can avoid unnecessary fetches
+        monsterPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterAttributes.name, questDict[QuestSecondMonsterKey]];
+        quest.secondMonster = (Monster *)[self getUniqueEntityWithEntityName:[Monster entityName] withPredicate:monsterPredicate];
+        if (quest.secondMonster) {
+            monsterPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterAttributes.name, questDict[QuestThirdMonsterKey]];
+            quest.thirdMonster = (Monster *)[self getUniqueEntityWithEntityName:[Monster entityName] withPredicate:monsterPredicate];
+            if (quest.thirdMonster) {
+                monsterPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterAttributes.name, questDict[QuestFourthMonsterKey]];
```
vokal-isaac, line 515: It may be worth creating a method on, say, the `Monster` class along the lines of `predicateForMonsterWithName:(NSString *)name` to make these predicate-creation lines shorter and less repetitive.

vokal-isaac, line 515: ... or maybe better, a method like `getMonsterWithName:(NSString *)name` or something.

MH4U Dex/Model/Private/_Monster.h
```diff
@@ -17,10 +17,18 @@ extern const struct MonsterAttributes {
...
+	__unsafe_unretained NSString *questSecondTarget;
+	__unsafe_unretained NSString *questThirdTarget;
 } MonsterRelationships;
 
 @class DamageZone;
 @class MonsterDrop;
+@class Quest;
+@class Quest;
+@class Quest;
+@class Quest;
```
vokal-isaac, line 31: <s>Is there a reason this is repeated four times?</s>  ... oh, nevermind, this is mogenerator-generated.

MH4U Dex/CoreDataController.m
```diff
@@ -177,6 +210,50 @@ - (void)loadMonsterDropData
...
+
+- (void)loadQuestDropData
+{
+    NSArray *questDropList = [self loadArrayFromJsonFileNamed:QuestDropsFileName];
+    
+    for (NSDictionary *dropDict in questDropList) {
+        QuestDrop *drop;
+        NSNumber *dropID = dropDict[QuestDropIDKey];
+        NSPredicate *dropPredicate = [NSPredicate predicateWithFormat:@"%K == %d", QuestDropAttributes.id, dropID.intValue];
+        drop = (QuestDrop *)[self getUniqueEntityWithEntityName:[QuestDrop entityName] withPredicate:dropPredicate];
```
vokal-isaac: You could combine the `QuestDrop *drop;` line with this line.

MH4U Dex/CoreDataController.m
```diff
@@ -177,6 +210,50 @@ - (void)loadMonsterDropData
...
+        if (!drop) {
+            drop = [self questDropFromDictionary:dropDict];
+            if (!drop) {
+                //If we reach this point, then the quest that this drop belongs to did not exist, and we should return without saving the context;
```
vokal-isaac, line 244: If I'm reading it correctly, it's not just that it did not exist, it's that you tried to create it and failed to do so...

MH4U Dex/CoreDataController.m
```diff
@@ -363,6 +439,87 @@ - (NSUInteger)countForEntityWithEntityName:(NSString *)entityName withPredicate:
...
+- (QuestDrop *)questDropFromDictionary:(NSDictionary *)dropDict
+{
+    QuestDrop *drop = [QuestDrop insertInManagedObjectContext:self.managedObjectContext];
+    NSPredicate *questPredicate = [NSPredicate predicateWithFormat:@"%K == %@", QuestAttributes.id, dropDict[QuestDropQuestIDKey]];
+    drop.quest = (Quest *)[self getUniqueEntityWithEntityName:[Quest entityName] withPredicate:questPredicate];
+    if (!drop.quest) {
+        //If the quest was not found, then it makes no sense to store this reward.  Just return nil.
+        NSLog(@"The quest for a quest reward was not found!");
+        return nil;
+    }
```
vokal-isaac, line 449: You will still have created the `QuestDrop` object `drop`, though—you might want to try to retrieve the quest first, check that it exists, then create `drop`.

MH4U Dex/QuestEncyclopediaContainerViewController.m
```diff
@@ -0,0 +1,38 @@
...
+@property (nonatomic, weak) IBOutlet UIView *hallSubView;
+
+@end
+
+@implementation QuestEncyclopediaContainerViewController
+
+- (IBAction)segmentedControlChanged:(id)sender
+{
+    
+}
```
vokal-isaac, line 25: Is this intentionally empty?  Should there be a `// TODO:` comment in here?

MH4U Dex/QuestEncyclopediaContainerViewController.m
```diff
@@ -0,0 +1,38 @@
...
+
+- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
+{
+    if ([segue.identifier isEqualToString:@"showCaravanList"]) {
+        QuestListTableViewController *questListVC = (QuestListTableViewController *)segue.destinationViewController;
+        questListVC.isCaravan = YES;
+    } else if ([segue.identifier isEqualToString:@"showGatheringHallList"]) {
+        QuestListTableViewController *questListVC = (QuestListTableViewController *)segue.destinationViewController;
```
vokal-isaac: This line is in both blocks above—would it make sense to pull it above the `if` statement?

MH4U Dex/QuestListTableViewController.m
```diff
@@ -0,0 +1,89 @@
...
+@end
+
+@implementation QuestListTableViewController
+
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    
+    
+
```
vokal-isaac: Extra blank lines.

MH4U Dex/QuestListTableViewController.m
```diff
@@ -0,0 +1,89 @@
...
+@implementation QuestListTableViewController
+
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    
+    
+    
+    CoreDataController *coreDataController = [CoreDataController sharedCDController];
```
vokal-isaac: It looks like you're only using this for the `.managedObjectContext` property—if so, it might make sense to have an `NSManagedObjectContext` variable and store the value of that property in it straight away, rather than storing the `CoreDataController`.

MH4U Dex/QuestListTableViewController.m
```diff
@@ -0,0 +1,89 @@
...
+    NSPredicate *caravanPredicate = [NSPredicate predicateWithFormat:@"%K == %@", QuestAttributes.caravan, [NSNumber numberWithBool:self.isCaravan]];
+    NSFetchRequest *questFetchRequest = [NSFetchRequest fetchRequestWithEntityName:[Quest entityName]];
+    self.questMasterArray = [NSMutableArray array];
+    NSPredicate *starPredicate;
+    NSCompoundPredicate *compoundPredicate;
+    for (NSUInteger starNumber = 1; starNumber <= 10; starNumber++) {
+        starPredicate = [NSPredicate predicateWithFormat:@"%K == %@", QuestAttributes.stars, @(starNumber)];
+        compoundPredicate = [NSCompoundPredicate andPredicateWithSubpredicates:@[caravanPredicate, starPredicate]];
+        [questFetchRequest setPredicate:compoundPredicate];
+        [self.questMasterArray addObject:[coreDataController.managedObjectContext executeFetchRequest:questFetchRequest error:&fetchError]];;
```
vokal-isaac: `executeFetchRequest:error:` might return `nil`, which would cause a crash for attempting to insert `nil` into an array.  You might also want to do something with `fetchError` when it returns `nil`.

vokal-isaac: Extra `;`.

MH4U Dex/RegionEncyclopediaCollectionViewController.m
```diff
@@ -8,12 +8,17 @@
...
 
+#import "CoreDataController.h"
+
+#import "RegionContainerViewController.h"
+
 #import "Region.h"
 
+#import "RegionEncylcopediaCollectionViewCell.h"
+
+
```
vokal-isaac, line 20: Extra blank line.

MH4U Dex/CoreDataController.m
```diff
@@ -93,33 +131,33 @@ - (void)loadMonsterData
...
+                   @"cut": @"cut",
+                   @"impact": @"impact",
+                   @"shot": @"shot",
+                   @"fire": @"fire",
+                   @"water": @"water",
+                   @"thunder": @"thunder",
+                   @"ice": @"ice",
+                   @"dragon": @"dragon",
+                   @"ko": @"ko",
+                   @"extract": @"extract",
```
vokal-isaac: Do constants exist for any of these strings?  Perhaps mogenerator-generated attribute names?

MH4U Dex/RegionEncylcopediaCollectionViewCell.m
```diff
@@ -8,6 +8,26 @@
...
+@property (nonatomic, weak) IBOutlet UILabel *regionNameLabel;
+@property (nonatomic, weak) IBOutlet UIImageView *regionIconImageView;
+
+@end
+
 @implementation RegionEncylcopediaCollectionViewCell
 
+- (void)displayContents
+{
+    NSString *imageName = [NSString stringWithFormat:@"%@-icon", self.regionKeyName];
```
vokal-isaac:
``` objc
NSString *imageName = [self.regionKeyName stringByAppendingString:@"-icon"];
```

Also, you've got basically this same construct in two far-apart places, so it might make sense to consolidate this logic in one more-centralized place.

MH4U Dex/MonsterEncyclopediaViewController.m
```diff
@@ -37,8 +39,8 @@ - (void)viewDidLoad
 
     NSError *fetchError = nil;
     NSFetchRequest *fetchRequest = [[NSFetchRequest alloc] initWithEntityName:[Monster entityName]];
-    NSSortDescriptor *sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:@"sort_name" ascending:YES];
-    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"%K == %@", @"monster_class", @"Minion"];
+    NSSortDescriptor *sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:MonsterAttributes.sort_name ascending:YES];
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterAttributes.monster_class, @"Minion"];
```
vokal-isaac: Are there constants for `@"Minion"` here and `@"Boss"` below?

MH4U Dex/RegionItemTableViewController.m
```diff
@@ -0,0 +1,107 @@
...
+    NSError *fetchError = nil;
+    NSSortDescriptor *areaSortDescriptor = [NSSortDescriptor sortDescriptorWithKey:AreaAttributes.name ascending:YES];
+    NSFetchRequest *areaFetchRequest = [[NSFetchRequest alloc] initWithEntityName:[Area entityName]];
+    NSPredicate *areaPredicate = [NSPredicate predicateWithFormat:@"%K == %@", AreaRelationships.region, self.region];
+    NSPredicate *rankPredicate;
+    if (self.rank) {
+        rankPredicate = [NSPredicate predicateWithFormat:@"%K == %@", AreaAttributes.rank, self.rank];
+    } else {
+        rankPredicate = [NSPredicate predicateWithFormat:@"%K == %@", AreaAttributes.rank, @"Low"];
+    }
```
vokal-isaac: I think I've seen these 5 lines somewhere else...

MH4U Dex/MonsterItemTableViewController.h
```diff
@@ -8,9 +8,10 @@
 
 #import <UIKit/UIKit.h>
 
-@interface MonsterLowRankTableViewController : UITableViewController
+@interface MonsterItemTableViewController : UITableViewController
 
 @property (nonatomic, strong) NSManagedObjectContext *managedObjectContext;
 @property (nonatomic, strong) NSString *monster;
```
vokal-isaac: I wouldn't expect a property named `monster` to be a string—should this be a `Monster` object or should it be called `monsterName` or something else?

MH4U Dex/MonsterOverviewViewController.h
```diff
@@ -13,6 +13,4 @@
 @property (nonatomic, strong) NSManagedObjectContext *managedObjectContext;
 @property (nonatomic, strong) NSString *monster;
```
vokal-isaac: Ditto here.

MH4U Dex/DropTableViewCell.h
```diff
@@ -15,6 +15,9 @@
 @property (nonatomic, assign) NSInteger percent;
 @property (nonatomic, assign) NSInteger quantity;
 
+/**
+ Update the cell's labels.
+ */
 - (void)displayContents;
```
vokal-isaac: Should this method be passed a model object?

MH4U Dex/ItemMonsterSourceTableViewCell.h
```diff
@@ -13,6 +13,9 @@
 @property (nonatomic, strong) NSString *method;
 @property (nonatomic, strong) NSString *sourceName;
 
+/**
+ Update the cell's labels.
+ */
 - (void)displayContents;
```
vokal-isaac: Should this method be passed a model object?

MH4U Dex/ItemTableViewCell.h
```diff
@@ -12,6 +12,10 @@
 
 @property (nonatomic, strong) NSString *itemName;
 
+
+/**
+ Update the cell's labels.
+ */
 - (void)displayContents;
```
vokal-isaac: Should this method be passed a model object?

MH4U Dex/MonsterEncyclopediaCell.h
```diff
@@ -12,6 +12,9 @@
 
 @property (nonatomic, strong) NSString *monsterName;
 
+/**
+ Update the cell's labels.
+ */
 - (void)displayContents;
```
vokal-isaac: Should this method be passed a model object?

MH4U Dex/QuestListTableViewCell.m
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Joseph Goldberg on 3/17/15.
+//  Copyright (c) 2015 Joseph Goldberg. All rights reserved.
+//
+
+#import "QuestListTableViewCell.h"
+
+@implementation QuestListTableViewCell
+
+@end
```
vokal-isaac, line 14: Empty class?  Should there be a `// TODO:` comment?

MH4U Dex/RegionEncylcopediaCollectionViewCell.h
```diff
@@ -10,4 +10,12 @@
 
 @interface RegionEncylcopediaCollectionViewCell : UICollectionViewCell
 
+@property (nonatomic, strong) NSString *regionName;
+@property (nonatomic, strong) NSString *regionKeyName;
+
+/**
+ Update the cell's labels.
+ */
+- (void)displayContents;
```
vokal-isaac: Should this method be passed a model object?

MH4U Dex/MonsterDetailsViewController.m
jmgoldb: How would I do this without a segmented control to pass as a parameter?

vokal-isaac: Have an `IBOutlet` pointed at the segmented control and pad that property as the sender.

MH4U Dex/CoreDataController.m
jmgoldb, line 471: It's related to how I was/am generating the data, it's easier and faster for me to generate it that way.

MH4U Dex/RegionEncylcopediaCollectionViewCell.h
jmgoldb: I could pass it a model object instead of storing the two strings as properties, if that would be preferable

seanwolter: Our preferred pattern is to make a method like `layoutWithThing:` or `configureWithThing:` and pass, but not retain, the model object.

MH4U Dex/CoreDataController.m
jmgoldb, line 244: I think you might be reading this incorrectly, let's talk about it later today.

MH4U Dex/MonsterDetailsViewController.m
jmgoldb: Changed these, but setting self.navigationItem.title has the same result as setting self.title.  I'm not sure why that is.

jmgoldb:
I've addressed all of the issues, aside from declaring constants for strings that get used more than once, which I'd like to do by making an external constants file as the next pull request.  I also haven't consolidated the image handling into one place, I'm not sure yet where I want to put that.  Everything else should be accounted for.

MH4U Dex/CoreDataController.m
```diff
@@ -363,6 +438,87 @@ - (NSUInteger)countForEntityWithEntityName:(NSString *)entityName withPredicate:
     }
 }
 
+- (QuestDrop *)questDropFromDictionary:(NSDictionary *)dropDict
+{
+    NSPredicate *questPredicate = [NSPredicate predicateWithFormat:@"%K == %@", QuestAttributes.id, dropDict[QuestDropQuestIDKey]];
+    if (![self getUniqueEntityWithEntityName:[Quest entityName] withPredicate:questPredicate]) {
```
vokal-isaac: To avoid having to run this `getUniqueEntityWithEntityName:withPredicate:` call twice, you could store it into a local variable, e.g.:

``` objc
Quest *quest = (Quest *)[self getUniqueEntityWithEntityName:[Quest entityName] withPredicate:questPredicate];
if (!quest) {
...
drop.quest = quest.
```

MH4U Dex/MonsterEncyclopediaViewController.m
```diff
@@ -62,9 +64,26 @@ - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
...
+{
+    switch (indexPath.section) {
+        case SmallMonsterSection:
+            return self.smallMonsters[indexPath.row];
+            break;
+        case LargeMonsterSection:
+            return self.largeMonsters[indexPath.row];
+            break;
+    }
+    return nil;
```
vokal-isaac: You don't need the `break;`s after `return`s, but you can leave them in if you want.  Similarly, it's a little more typical to include a `default:` section and put the `return nil;` in there.

MH4U Dex/MonsterDetailsViewController.m
```diff
@@ -29,43 +35,57 @@ @implementation MonsterDetailsViewController
...
+    
+    if ([segue.identifier isEqualToString:@"embedOverview"]) {
         MonsterOverviewViewController *overViewVC = (MonsterOverviewViewController *)segue.destinationViewController;
         overViewVC.managedObjectContext = self.managedObjectContext;
-        overViewVC.monster = self.monster;
+        overViewVC.monsterName = self.monster;
+    } else {
+        MonsterItemTableViewController *itemVC = (MonsterItemTableViewController *)segue.destinationViewController;
+        itemVC.managedObjectContext = self.managedObjectContext;
+        itemVC.monsterName = self.monster;
```
vokal-isaac: Is this (and 4 lines farther up) another instance of a property named `monster` that's a name string rather than a `Monster` object?

MH4U Dex/CoreDataController.m
seanwolter: I know I'm late to the game on this project. Please forgive these points:
1) Generally "get" as a prefix is verboten. I'd prefer "fetch" or no prefix for a method like this.
2) Why not use Vokoder? I respect the experienced gained building your own core data stack, but I'd also like you to be up to speed on our standard tools.

MH4U Dex/QuestListTableViewController.m
```diff
@@ -0,0 +1,92 @@
...
+    NSManagedObjectContext *managedObjectContext = [CoreDataController sharedCDController].managedObjectContext;
+    NSError *fetchError = nil;
+    NSPredicate *caravanPredicate = [NSPredicate predicateWithFormat:@"%K == %@", QuestAttributes.caravan, [NSNumber numberWithBool:self.isCaravan]];
+    NSFetchRequest *questFetchRequest = [NSFetchRequest fetchRequestWithEntityName:[Quest entityName]];
+    self.questMasterArray = [NSMutableArray array];
+    NSPredicate *starPredicate;
+    NSCompoundPredicate *compoundPredicate;
+    for (NSUInteger starNumber = 1; starNumber <= 10; starNumber++) {
+        starPredicate = [NSPredicate predicateWithFormat:@"%K == %@", QuestAttributes.stars, @(starNumber)];
+        compoundPredicate = [NSCompoundPredicate andPredicateWithSubpredicates:@[caravanPredicate, starPredicate]];
```
vokal-isaac: Can this line (and `NSCompoundPredicate *compoundPredicate;` above) be removed?

MH4U Dex/RegionEncyclopediaCollectionViewController.m
```diff
@@ -53,13 +56,34 @@ - (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSe
...
+    if ([segue.identifier isEqualToString:@"showRegionDetails"]) {
+        RegionContainerViewController *regionVC = (RegionContainerViewController *)segue.destinationViewController;
+        regionVC.managedObjectContext = self.managedObjectContext;
+        
+        if ([sender isMemberOfClass:[RegionEncyclopediaCollectionViewCell class]]) {
+            RegionEncyclopediaCollectionViewCell *cell = (RegionEncyclopediaCollectionViewCell *)sender;
+            Region *region = [self regionAtIndexPath:[self.collectionView indexPathForCell:cell]];
+            regionVC.regionName = region.name;
+            regionVC.regionKeyName = region.keyName;
+            regionVC.region = region;
```
vokal-isaac, line 82: If you're setting a `region` property, are the `regionName` and `regionKeyName` properties on `regionVC` necessary?

vokal-isaac:
Good to pull.

Keep at least a mental TODO to look at switching to Vokoder, as Sean suggested.

seanwolter:
Joe answered my question about Vokoder vs. build a core data castle from scratch.
Were the `get...` methods changed? Or does no one else mind that?

vokal-isaac:
I believe d8a5f0b renamed the `get...` method.

PR #5, opened 2015-03-20
Monster and MonsterDrop Overhaul

jmgoldb:
Monster and MonsterDrop loading have both been overhauled, and both function correctly!  Monster Damage Zones are next to be overhauled, so I've bypassed hooking them back up for now.  The app just doesn't load in the damage zones at the moment.


MH4U Dex/CoreDataController.h
```diff
@@ -14,34 +14,19 @@
...
-/**
- Load into Core Data the rewards that can be obtained from Quests
- */
-- (void)loadQuestDropData;
-
-/**
- Attempt to save the persistent store, and return true if succesful
- */
-- (BOOL)attemptSaveContext;
++ (BOOL)willReloadUponRestart;
```
vokal-isaac: If this is matching the previous `set...` method, I'm thinking it should be `shouldTriggerReloadUponRestart`.

MH4U Dex/CoreDataController.m
```diff
@@ -46,88 +48,79 @@ + (CoreDataController *)sharedCDController
...
-            newDrop.quantity = drop[MHDMonsterDropQuantityKey];
-            
-            // If the Item for this drop already exists in the persistent store ...
-            NSPredicate *itemPredicate = [NSPredicate predicateWithFormat:@"%K == %@", ItemAttributes.name, drop[MHDMonsterDropItemNameKey]];
-            Item *item = (Item *)[self uniqueEntityWithEntityName:[Item entityName] withPredicate:itemPredicate];
-            if (!item) {
-                item = [Item insertInManagedObjectContext:self.managedObjectContext];
-                item.name = drop[MHDMonsterDropItemNameKey];
+    //Repeat this process for ten different JSON files
+    for (NSUInteger fileNumber = 1; fileNumber <= 10; fileNumber++) {
```
vokal-isaac: Does the `1` through `10` have to be hard-coded?  Could you find all files matching a pattern and iterate through them instead?  Or something else?

MH4U Dex/CoreDataController.m
```diff
@@ -46,88 +48,79 @@ + (CoreDataController *)sharedCDController
...
-                if (!monster) {
-                    // TODO: Handle better, later.
-                    NSLog(@"The Monster, %@, was not found!", monsterName);
-                } else {
-                    newDrop.monsterSource = monster;
-                    newDrop.rank = drop[MHDMonsterDropRankKey];
-                    newDrop.percent = drop[MHDMonsterDropPercentKey];
-                }
+            NSPredicate *monsterDropPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropAttributes.id, monsterDropDict[MHDMonsterDropIDKey]];
+            MonsterDrop *monsterDrop = (MonsterDrop *)[self uniqueEntityWithEntityName:[MonsterDrop entityName] withPredicate:monsterDropPredicate];
```
vokal-isaac: Consider re-breaking these two lines so they aren't as long (lines should generally be 120 characters or fewer).

MH4U Dex/CoreDataController.m
```diff
@@ -441,6 +434,38 @@ - (Quest *)questFromDictionary:(NSDictionary *)questDict
...
+    NSPredicate *itemPredicate = [NSPredicate predicateWithFormat:@"%K == %@", ItemAttributes.name, name];
+    Item *item = (Item *)[self uniqueEntityWithEntityName:[Item entityName] withPredicate:itemPredicate];
+    if (!item) {
+        item = [Item insertInManagedObjectContext:self.managedObjectContext];
+        item.name = name;
+    }
+    return item;
+}
+
+#pragma makr - Core Data Helper Methods
```
vokal-isaac: s/makr/mark/

MH4U Dex/ItemMonsterSourcesTableViewController.m
```diff
@@ -74,9 +77,8 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
 - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
 {
     UIStoryboard *storyBoard = [UIStoryboard storyboardWithName:@"Main" bundle:nil];
-    MonsterDetailsViewController *monsterVC = [storyBoard instantiateViewControllerWithIdentifier:@"MonsterDetailsViewController"];
-    monsterVC.managedObjectContext = self.managedObjectContext;
-    monsterVC.monsterName = [self dropAtIndexPath:indexPath].monsterSource.name;
+    MonsterContainerViewController *monsterVC = [storyBoard instantiateViewControllerWithIdentifier:@"MonsterDetailsViewController"];
```
vokal-isaac: Would it make sense to replace the string literal here with something like `NSStringFromClass([MonsterDetailsViewController class])`?

MH4U Dex/MonsterItemTableViewController.m
```diff
@@ -32,27 +33,24 @@ @implementation MonsterItemTableViewController
...
+    NSManagedObjectContext *managedObjectContext = [CoreDataController sharedCDController].managedObjectContext;
     NSError *fetchError = nil;
-    NSFetchRequest *monsterFetchRequest = [[NSFetchRequest alloc] initWithEntityName:[Monster entityName]];
-    monsterFetchRequest.predicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterAttributes.name, self.monsterName];
-    Monster *monster = (Monster *)[self.managedObjectContext executeFetchRequest:monsterFetchRequest error:&fetchError].firstObject;
     
     NSFetchRequest *fetchRequest = [[NSFetchRequest alloc] initWithEntityName:[MonsterDrop entityName]];
-    NSSortDescriptor *sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:MonsterDropAttributes.method ascending:YES];
-    NSPredicate *dropPredicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropRelationships.monsterSource, monster];
+    NSSortDescriptor *sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:MonsterDropAttributes.id ascending:YES];
```
vokal-isaac, line 40: Is it intentional that you're sorting on the `id`?

vokal-isaac:
@chillpop @designatednerd @seanwolter Any of you have time to see if there's anything you want to comment on?

MH4U Dex/CoreDataController.m
jmgoldb: oops

MH4U Dex/ItemMonsterSourcesTableViewController.m
jmgoldb: Sure would

MH4U Dex/MonsterItemTableViewController.m
jmgoldb, line 40: Yes

MH4U Dex/CoreDataController.m
jmgoldb: I suppose I could do this another way, but is there a problem doing this like this?

I was planning on reusing this pattern for the quests, which are actually divided into numbers 1-10

chillpop:
nothing to add

MH4U Dex/CoreDataController.m
vokal-isaac: In general, hard-coded numbers are bad.  That doesn't mean every specific instance is inherently bad, but if the numbers correspond to file names and nothing else, it seems like you could avoid the hard-coded numbers by looking for files that match a pattern.

MH4U Dex/Base.lproj/Main.storyboard
```diff
@@ -156,6 +156,15 @@
                                     <segue destination="hfi-eK-gsg" kind="show" id="dX6-Hf-8Bu"/>
                                 </connections>
                             </button>
+                            <button opaque="NO" contentMode="scaleToFill" fixedFrame="YES" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="roundedRect" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="Cxa-Tg-f1x">
+                                <rect key="frame" x="35" y="377" width="136" height="30"/>
+                                <state key="normal" title="RESET CORE DATA">
```
designatednerd, line 161: Just a general note: make sure when you're setting titles of things in storyboards that you're also making sure you're using appropriate methods to localize - otherwise things may not get localized properly.

MH4U Dex/CoreDataController.m
```diff
@@ -46,88 +48,79 @@ + (CoreDataController *)sharedCDController
 
 #pragma mark - Initial Data Loading
 
+- (void)tryLoadSequence
+{
+    if ([[NSUserDefaults standardUserDefaults] boolForKey:NeedsReloadKey]) {
+        NSLog(@"Resetting Core Data");
+        [self resetCoreData];
```
designatednerd, line 57: Couple thoughts here: 

a) You will likely want to use `DLog` rather than `NSLog` to prevent log barf in the release version. 
b) If you need to log these things, would it make more sense to put those logs in the individual methods so that you can see every time they're getting called rather than just when they're called in this load sequence?

MH4U Dex/CoreDataController.m
designatednerd: Even if you're using static consts, those at least are more expressive - they tell you _what_ 1 and 10 are rather than just that the numbers are 1 and 10.

MH4U Dex/MonsterItemTableViewController.m
designatednerd, line 40: Is the identifier something you're controlling?

MH4U Dex/CoreDataController.m
jmgoldb, line 57: I should have put in a TODO:, but I actually want to remove those NSLog calls once I'm sure everything works as it should be.  I tried to keep them all in that place so I could remove them a bit more easily.

MH4U Dex/MonsterItemTableViewController.m
jmgoldb, line 40: Yeah, it's not randomly generated.  When I manually built the data in google spreadsheets, I assigned each row an auto-incrementing ID.  Sorting by ID is useful because it preserves the order I have in my source spreadsheets, which happens to be the order I'd like to use in the app anyways.

MH4U Dex/Base.lproj/Main.storyboard
jmgoldb, line 161: Thanks.  I've put a TODO: for that in the view controller's implementation file.  I might move or change the buttons so I won't hook them up to IBOutlets just yet.

MH4U Dex/CoreDataController.m
jmgoldb: What do you guys think I should use for this situation?  For the monsterDrops, like monsters are grouped up together within the same numbering, but otherwise the numbers are meaningless, and the files are only split up to avoid dealing with overly-large JSON files.  It sounds like I really should just look for files matching a pattern in this scenario, and maybe use static consts for the quests scenario?

vokal-isaac: As I mentioned in our offline discussion, the motivation to me is that it's preferable to avoid hard-coding numbers that are already represented elsewhere—that is, if it can be inferred from the set of files on disk that there are files numbered 1–10, it's better to make that inference rather than hard-coding the numbers because changing what files exist on disk will automatically change what's loaded without requiring a corresponding code change.

MH4U Dex/CoreDataController.m
vokal-isaac, line 57: There's also no `DLog` macro in here at the moment—the current Xcode default project template doesn't include a `.pch` file, so it'd have to be added manually in order to put a `DLog` macro in it.

vokal-isaac:
Good to pull.

PR #8, opened 2015-03-30
Item Encyclopedia has search and indexing.

jmgoldb:
Can I get another quick code review, please?  @vokal/ios-developers


brockboland:
on it

MH4U Dex/ItemEncyclopediaTableViewController.m
```diff
@@ -46,12 +42,12 @@ - (void)viewDidLoad
...
 - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView
 {
-    return 1;
+    return self.fetchedResultsController.sections.count;
 }
 
 - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section
 {
-    return self.items.count;
+    return [[self.fetchedResultsController.sections objectAtIndex:section] numberOfObjects];
```
brockboland: Since `sections` is an array, you could do this if you want to:

``` objc
return [self.fetchedResultsController.sections[section] numberOfObjects];
```

brockboland: Actually, `numberOfObjects` is a property, too, so this should work (but Xcode might complain because it doesn't know where `numberOfObjects` is supposed to come from):

``` objc
return self.fetchedResultsController.sections[section].numberOfObjects;
```

MH4U Dex/ItemEncyclopediaTableViewController.m
```diff
@@ -61,11 +57,39 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
     return cell;
 }
 
+- (NSArray *)sectionIndexTitlesForTableView:(UITableView *)tableView
+{
+    //return [[UILocalizedIndexedCollation currentCollation] sectionIndexTitles];
```
brockboland: Commented-out line here should be removed

MH4U Dex/ItemEncyclopediaTableViewController.m
```diff
@@ -61,11 +57,39 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
+}
+
+- (NSFetchedResultsController *)fetchedResultsControllerWithPredicate:(NSPredicate *)predicate
+{
+    NSError *fetchError = nil;
+    NSFetchRequest *fetchRequest = [[NSFetchRequest alloc] initWithEntityName:[Item entityName]];
+    fetchRequest.sortDescriptors = @[[NSSortDescriptor sortDescriptorWithKey:ItemAttributes.name ascending:YES]];
+    fetchRequest.fetchBatchSize = 20;//TODO: Tinker with this.
+    fetchRequest.predicate = predicate;
+    NSFetchedResultsController *newResultsController = [[NSFetchedResultsController alloc] initWithFetchRequest:fetchRequest managedObjectContext:[CoreDataController sharedCDController].managedObjectContext sectionNameKeyPath:@"firstLetter" cacheName:@"Root"];
```
brockboland: This line should be broken up, since it has 3 or more params in the method call.

MH4U Dex/ItemEncyclopediaTableViewController.m
```diff
@@ -81,4 +105,17 @@ - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
...
+#pragma mark - Search Bar Delegate Methods
+
+- (void)searchBar:(UISearchBar *)searchBar textDidChange:(NSString *)searchText
+{
+    if (!searchText.length) {
+        self.fetchedResultsController = [self fetchedResultsControllerWithPredicate:nil];
+    } else {
+        NSPredicate *predicate = [NSPredicate predicateWithFormat:@"%K contains[cd] %@", ItemAttributes.name, searchText];
+        self.fetchedResultsController = [self fetchedResultsControllerWithPredicate:predicate];
+    }
```
brockboland, line 120: This one is debatable, but I think I would change the `fetchedResultsControllerWithPredicate:` method to something like `fetchedResultsControllerWithSearchTerm:`, and make that method responsible for building the NSPredicate. That way, the predicate setup is centralized.

MH4U Dex/ItemEncyclopediaTableViewController.m
brockboland: It's usually a good idea to avoid hard-coding method or property names like "firstLetter" here. If you make the `firstLetter` method public, you could then use this for the `sectionNameKeyPath` param:

```
NSStringFromSelector(@selector(firstLetter))
```

MH4U Dex/MonsterEncyclopediaHeader.m
```diff
@@ -0,0 +1,28 @@
...
+@end
+
+@implementation MonsterEncyclopediaHeader
+
+- (void)configureHeaderLabelWithMonsterIsSmall:(BOOL)isSmall
+{
+    if (isSmall) {
+        self.headerLabel.text = @"Small Monsters";
+    } else {
+        self.headerLabel.text = @"Large Monsters";
```
brockboland: You should use `NSLocalizedString` for any string like this that's displayed to the user, but I'm not sure if you're doing so in this project.

MH4U Dex/MonsterEncyclopediaViewController.m
```diff
@@ -104,6 +105,20 @@ - (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSe
     }
 }
 
+- (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView viewForSupplementaryElementOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath
```
brockboland: Three params, should be broken into multiple lines

MH4U Dex/MonsterEncyclopediaViewController.m
```diff
@@ -104,6 +105,20 @@ - (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSe
...
+{
+    MonsterEncyclopediaHeader *reusableView = [collectionView dequeueReusableSupplementaryViewOfKind:kind withReuseIdentifier:NSStringFromClass([MonsterEncyclopediaHeader class]) forIndexPath:indexPath];
```
brockboland: Three params

brockboland:
You've got some commented-out code that should go and a few long lines to break up, but the other stuff is just suggestions: it's up to you whether you make those changes.

jmgoldb:
Thanks for the code review.  Pushed a new commit addressing most of the comments

brockboland:
Cool, good to pull

PR #11, opened 2015-03-31
Item source refactor

jmgoldb:
Going to an item in the item encylcopedia and viewing the sources of an item is much better, now.


jmgoldb:
Forgot to mention one of the JSON files had a search and replace change, so the linecount is really inflated.

jmgoldb:
Is anyone up for a code review?  @vokal/ios-developers

MH4U Dex/ItemSourceTableViewCell.m
```diff
@@ -0,0 +1,44 @@
...
+- (void)displayContentsWithDrop:(Drop *)drop
+{
+    if ([drop isKindOfClass:[MonsterDrop class]]) {
+        MonsterDrop *monsterDrop = (MonsterDrop *)drop;
+        self.methodLabel.text = monsterDrop.method;
+        self.sourceLabel.text = monsterDrop.monsterSource.name;
+        
+    } else if ([drop isKindOfClass:[AreaDrop class]]) {
+        AreaDrop *areaDrop = (AreaDrop *)drop;
+        self.methodLabel.text = areaDrop.method;
```
brockboland: You could move the `methodLabel` setting out of the conditional, since it's the same for both Drop types:

``` objc
self.methodLabel.text = drop.method;
```

MH4U Dex/ItemSourcesTableViewController.m
```diff
@@ -0,0 +1,112 @@
...
+@end
+
+@implementation ItemSourcesTableViewController
+
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    
+    NSManagedObjectContext *managedObjectContext = [CoreDataController sharedCDController].managedObjectContext;
+    NSError *fetchError = nil;
```
brockboland: I know you just moved this from another file, but: you don't need to explicitly set new variables to nil.

MH4U Dex/ItemSourcesTableViewController.m
```diff
@@ -0,0 +1,112 @@
...
+    NSFetchRequest *itemFetchRequest = [[NSFetchRequest alloc] initWithEntityName:[Item entityName]];
+    itemFetchRequest.predicate = [NSPredicate predicateWithFormat:@"%K == %@", ItemAttributes.name, self.itemName];
```
brockboland, line 43: Again, I know this is existing code, but: do you have a unique ID or something on Items that could be used instead of `itemName` here? It's generally a bad idea to key on names, unless the names are coming from a pre-defined (and ideally, hard-coded) list of options, to ensure you don't miss on something silly like "hammer" vs "Hammer."

MH4U Dex/ItemSourcesTableViewController.m
```diff
@@ -0,0 +1,112 @@
...
+        fetchRequest = [[NSFetchRequest alloc] initWithEntityName:[MonsterDrop entityName]];
+        fetchRequest.predicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropRelationships.item, item];
+        fetchRequest.sortDescriptors = @[[NSSortDescriptor sortDescriptorWithKey:MonsterDropAttributes.id ascending:YES]];
+        self.tableView.accessibilityIdentifier = MHDItemMonsterSources;
+    } else {
+        
+        fetchRequest = [[NSFetchRequest alloc] initWithEntityName:[AreaDrop entityName]];
+        fetchRequest.predicate = [NSPredicate predicateWithFormat:@"%K == %@", AreaDropRelationships.item, item];
+        fetchRequest.sortDescriptors = @[[NSSortDescriptor sortDescriptorWithKey:AreaDropAttributes.idDecimalString ascending:YES]];
+        self.tableView.accessibilityHint = MHDItemAreaSources;
```
brockboland: This should be `accessibilityIdentifier`, not `accessibilityHint`.

MH4U Dex/ItemSourcesTableViewController.m
```diff
@@ -0,0 +1,112 @@
...
+    if (self.isMonsterSource) {
+        fetchRequest = [[NSFetchRequest alloc] initWithEntityName:[MonsterDrop entityName]];
+        fetchRequest.predicate = [NSPredicate predicateWithFormat:@"%K == %@", MonsterDropRelationships.item, item];
+        fetchRequest.sortDescriptors = @[[NSSortDescriptor sortDescriptorWithKey:MonsterDropAttributes.id ascending:YES]];
+        self.tableView.accessibilityIdentifier = MHDItemMonsterSources;
+    } else {
+        
+        fetchRequest = [[NSFetchRequest alloc] initWithEntityName:[AreaDrop entityName]];
+        fetchRequest.predicate = [NSPredicate predicateWithFormat:@"%K == %@", AreaDropRelationships.item, item];
+        fetchRequest.sortDescriptors = @[[NSSortDescriptor sortDescriptorWithKey:AreaDropAttributes.idDecimalString ascending:YES]];
```
brockboland: If I'm reading this right, your `fetchRequest` shouldn't be necessary. It looks like you're fetching drops that have a relationship back to the Item object you have here. In CoreData, relationships should always be reciprocal: if the drop has a relationship to the Item, the Item should have one back to the drop(s). In that case, the drops would be available as `item.drops` or something of that ilk. Am I misunderstanding what's going on here?

MH4U Dex/ItemSourcesTableViewController.m
```diff
@@ -0,0 +1,112 @@
...
+    return 1;
+}
+
+- (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section
+{
+    if (self.sources.count) {
+        return self.sources.count;
+    } else {
+        return 0;
+    }
```
brockboland: You can just return `self.sources.count` here without the conditional: if `self.sources` isn't an array object, it will return 0 anyway.

MH4U Dex/ItemSourcesTableViewController.m
```diff
@@ -0,0 +1,112 @@
...
+}
+
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    ItemSourceTableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:NSStringFromClass([ItemSourceTableViewCell class])];
+    [cell displayContentsWithDrop:self.sources[indexPath.row]];
```
brockboland: Any reason this doesn't call `dropAtIndexPath:` to get the object that's passed to `displayContentsWithDrop:`?

MH4U Dex/ItemSourceTableViewCell.m
jmgoldb: There will eventually be a third drop type that doesn't have a method, so method isn't a property of the abstract class.

MH4U Dex/ItemSourcesTableViewController.m
jmgoldb, line 43: The names actually are all pre-defined, and every item has a unique name.  It IS possible that I might get that situation due to a typo, I could probably write a test for that if it's necessary.

MH4U Dex/ItemSourcesTableViewController.m
jmgoldb: You might be right!  I need to convert it from a Set to an Array, though, and I want to keep the sort descriptor.  I'll look into it.

jmgoldb: Looks like sortedArrayUsingDescriptors will do the trick

MH4U Dex/ItemSourceTableViewCell.m
brockboland: Ah, ok

MH4U Dex/ItemSourcesTableViewController.m
brockboland, line 43: As long as those strings are only used in one place—as in, you use some const or struct every time you need to reference them—that should be fine.

jmgoldb:
Pushed some changes.  Think you could take another look once Travis finishes?

brockboland:
Looks good to me! Just need to wait for the green light from Travis.

seanwolter:
Neato! Travis!

PR #12, opened 2015-04-01
Blank page refactor

jmgoldb:
When tableViews come up empty, they now display messages to that effect.

### DIFF
--- a/MH4U Dex.xcodeproj/project.pbxproj
+++ b/MH4U Dex.xcodeproj/project.pbxproj
@@ -56,6 +56,7 @@
 		E4456EB61A89664F00FCC7DC /* DamageZone.m in Sources */ = {isa = PBXBuildFile; fileRef = E4456EB51A89664F00FCC7DC /* DamageZone.m */; };
 		E45E9A8E1A890CC400D75592 /* MonsterContainerViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = E45E9A8D1A890CC400D75592 /* MonsterContainerViewController.m */; };
 		E45E9A911A890ED300D75592 /* MonsterItemTableViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = E45E9A901A890ED300D75592 /* MonsterItemTableViewController.m */; };
+		E45F51A61ACC648E00C65741 /* UITableViewController+HeaderFooterChanger.m in Sources */ = {isa = PBXBuildFile; fileRef = E45F51A51ACC648E00C65741 /* UITableViewController+HeaderFooterChanger.m */; };
 		E46BBBA31AAA483100068F02 /* RegionContainerViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = E46BBBA21AAA483000068F02 /* RegionContainerViewController.m */; };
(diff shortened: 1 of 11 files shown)


jmgoldb:
Can I get a code review, please? @vokal/ios-developers

brockboland:
Sure

jmgoldb:
Thanks!

brockboland:
This is one of those cases where I'm not sure if this is the best course of action, but only because I'm not looking at the full app. Does the `tableHeaderView` typically have a different background color than the `tableView` itself? Are the `scrollEnabled` and `separatorColor` sets just to get rid of those lines that appear kind of like placeholders for cells that don't exist? (if so, the simplest solution to that is to add an empty `tableFooterView` on the table view: http://stackoverflow.com/a/14461000/2185)

Also, it looks like you're reaching the point where it would make sense to refactor shared code up into a common ancestor. Right now, these view controllers are all children of `UITableViewController`; I think it would make sense to add a `UITableViewController` subclass that these three can then sit below. Right now, it looks like that class would only need something like a `hideTableIfEmpty` method, which would check the number of cells being displayed—`tableView:numberOfRowsInSection:` for section 0 might do the trick, but I'm not sure about the `RegionItemTableViewController` because it can have multiple sections. In any case, the `hideTableIfEmpty` method would be called at the end of your `viewDidLoad` methods, in place of these new conditionals you added. What do you think?

jmgoldb:
Let me take a look at how that empty tableFooterView trick works out, and I'll get back to you.  And I'll also take a look at what refactoring the tableViews to a common ancestor would look like.

That ancestor would probably be an abstract class, right?

jmgoldb:
The empty tableFooterView works great, so I changed those.  I also did a tiny bit of refactoring where instead of passing around an itemName string, I just pass around the item directly.  I'm going to push what I have, and then take a look at refactoring into a common ancestor.

jmgoldb:
After looking at it some more, I'm not so sure I should bring everything into a common ancestor other than UITableViewController, because the tableViewControllers i have work with data in different ways.  MonsterItemTableViewController is fairly bog-standard, since it only holds an array of one type of object (MonsterDrop) but ItemSourceTableViewController has to deal with two different types of model objects, and RegionItemTableViewController has to worry about both the Areas in a region as well as the AreaDrops within each of those areas.

I could probably consolidate all of those into a base class anyways, but I'm not sure how much code reuse it will prevent.

I suppose I could refactor all of them to work with NSFetchedResultsController like the ItemEncyclopediaTableViewController, but I think that might be outside the scope of this pull request, and should probably be in its own pull request.  What do you think?

brockboland:
I was only suggesting that the one method be moved up to the parent class, not the rest of it, though you may well notice other functionality that could be refactored up there later on. The parent class would be "abstract" insomuch as you won't ever use it on its own, though Objective-C doesn't have truly abstract classes. Actually, got that reason, it would make just as much sense to create a category for `UITableViewController` and put the method in there.

jmgoldb:
A category sounds like a great suggestion!

I pushed things out to a category, though I'm not sure if I named it correctly or not.  How's it look?

jmgoldb:
Ah shoot, I pushed to the wrong branch.  Hang on

jmgoldb:
Okay, pushed things to the right branch this time.

brockboland:
cool, good to pull

jmgoldb:
Thanks!

PR #16, opened 2015-04-07
Progress hud and segmented control tweaks

jmgoldb:
Hey @vokal-isaac , can you tell me what you think of these changes?


MH4U Dex/HomeViewController.m
```diff
@@ -28,6 +30,21 @@ - (IBAction)resetButtonPressed:(id)sender
...
 }
 
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    
+    [MBProgressHUD showHUDAddedTo:self.view animated:YES];
+    dispatch_queue_t queue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
+    dispatch_async(queue, ^{
+        [[CoreDataController sharedCDController] tryLoadSequence];
```
vokal-isaac, line 40: Is this using a managed object context created on the main thread?

MH4U Dex/HomeViewController.m
```diff
@@ -28,6 +30,21 @@ - (IBAction)resetButtonPressed:(id)sender
...
+        dispatch_async(dispatch_get_main_queue(), ^{
```
vokal-isaac: Usually, it's (very slightly) preferable to use

``` objc
[[NSOperationQueue mainQueue] addOperationWithBlock:^{
    ...
}];
```

(because of a slight general preference for higher-level/Objective-C mechanisms over lower-level/pure-C mechanisms when both are available); there isn't a direct analog for the background-dispatch call, though.

vokal-isaac:
In the future, try to keep pods changes separate from non-pods changes to make reviewing easier.

Beyond that, really just the one question about Core Data.

MH4U Dex/HomeViewController.m
jmgoldb, line 40: I believe so.  My singleton CoreDataController only ever has/uses one managedObjectContext, and the getter for that doesn't specify anything about a separate thread.  Is there some weird interaction between threads and the managed object context that I'm not aware of?

jmgoldb:
Sorry, I totally forgot how crazy adding a pod would make the changelog

MH4U Dex/HomeViewController.m
vokal-isaac, line 40: It's not safe to use an `NSManagedObjectContext` instance across multiple threads and probably not safe to use it on a thread other than the one on which it was created.

jmgoldb, line 40: If that's the case, then what do you suggest I should do?  In the best case scenario, where a prebuilt-store is being used, the only time managedObjectContext is being called here is when it is being set to nil.  But if a prebuilt store isn't being used, then all of the various JSON-loading methods can potentially get called by tryLoadSequence.   How would I work with all of those in a background-thread context?  And would I still be able to access them later with a main-thread context?

jmgoldb, line 40: Should I make a backgroundObjectContext property and, when it is created, create it in a dispatch_async block?  The issue with that is, how would I then limit it to one specific background thread?

vokal-isaac, line 40: I think the best mechanism would be to call `tryLoadSequence` on the main thread and have it dispatch to the background if it's loading from JSON.

To do the loading from JSON, create a separate `NSManagedObjectContext` instance on the background thread, using the same persistent store; do all the loading with that managed object context; save it; then switch back to the main thread and make the shared context sync up with the persistent store.

jmgoldb:
@vokal-isaac I think I've fixed the Core Data issues, I've tested on my machine both with prebuilt and no prebuilt stores, and things SEEM to work fine, but If you could take another look I'd really appreciate it

vokal-isaac:
The changes look good to me, but Travis doesn't like it.

jmgoldb:
Ah shoot, let me investigate why.

jmgoldb:
Alright, travis is green

jmgoldb:
@vokal/ios-developers  Can I get a re-review and/or a merge, please?

seanwolter:
:suspect:

seanwolter:
Good to pull

PR #19, opened 2015-04-13
Quest Details View Implemented

jmgoldb:
This one seemed a little bit sprawling behind the scenes, though all the functionality is working exactly as intended.  Can I get a code review, please?

### DIFF
--- a/MH4U Dex/QuestContainerViewController.m
+++ b/MH4U Dex/QuestContainerViewController.m
@@ -8,6 +8,8 @@
 
 #import "QuestContainerViewController.h"
 
+#import "Constants.h"
+
 #import "Quest.h"
 
 #import "QuestDetailsTableViewController.h"
@@ -52,7 +54,7 @@ typedef NS_ENUM(NSInteger, QuestContainerSegmentedControlPage) {
 
 - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
 {
-    if ([segue.identifier isEqualToString:@"embedQuestDetails"]) {
+    if ([segue.identifier isEqualToString:MHDSegueIdentifier.embedQuestDetails]) {
         QuestDetailsTableViewController *detailVC = (QuestDetailsTableViewController *)segue.destinationViewController;
         detailVC.quest = self.quest;
     }
--- a/MH4U Dex/QuestContainerViewController.m
+++ b/MH4U Dex/QuestContainerViewController.m
@@ -1,0 +1,61 @@
+//
+//  QuestContainerViewController.m
+//  MH4U Dex
+//
+//  Created by Joseph Goldberg on 4/13/15.
+//  Copyright (c) 2015 Joseph Goldberg. All rights reserved.
+//
+
+#import "QuestContainerViewController.h"
+
+#import "Quest.h"
+
+#import "QuestDetailsTableViewController.h"
+
+typedef NS_ENUM(NSInteger, QuestContainerSegmentedControlPage) {
+    Details = 0,
+    Rewards = 1,
+    
+    SegmentedControlPageCount
+};
+
+@interface QuestContainerViewController ()
+
+@property (nonatomic, weak) IBOutlet UIView *detailsSubView;
+@property (nonatomic, weak) IBOutlet UIView *rewardsSubView;
+
+@end
+
+@implementation QuestContainerViewController
+
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    
+    self.navigationItem.title = self.quest.name;
+    self.rewardsSubView.hidden = YES;
+}
+
+- (IBAction)segmentedControlChanged:(UISegmentedControl *)segmentedControl
+{
+    switch (segmentedControl.selectedSegmentIndex) {
+        case Details:
+            self.detailsSubView.hidden = NO;
+            self.rewardsSubView.hidden = YES;
+            break;
+        case Rewards:
+            self.detailsSubView.hidden = YES;
+            self.rewardsSubView.hidden = NO;
+            break;
+    }
+}
+
+- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
+{
+    if ([segue.identifier isEqualToString:@"embedQuestDetails"]) {
+        QuestDetailsTableViewController *detailVC = (QuestDetailsTableViewController *)segue.destinationViewController;
+        detailVC.quest = self.quest;
+    }
+}
+
+@end
--- a/MH4U Dex/QuestDetailsTableViewController.m
+++ b/MH4U Dex/QuestDetailsTableViewController.m
@@ -76,6 +76,7 @@ static NSString *PointsConst = @"HRP Reward";
 
 - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView
 {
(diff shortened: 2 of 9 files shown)


jmgoldb:
@vokal/ios-developers Specifically, I'm not sure if it's a good idea to be using nested switch statements the way that I am.

brockboland:
Let's have a look-see here

MH4U Dex/QuestContainerViewController.m
```diff
@@ -0,0 +1,61 @@
...
+        case Rewards:
+            self.detailsSubView.hidden = YES;
+            self.rewardsSubView.hidden = NO;
+            break;
+    }
+}
+
+- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
+{
+    if ([segue.identifier isEqualToString:@"embedQuestDetails"]) {
```
brockboland: You can get away with it here, but it's generally a good to put storyboard identifies in constants; structs are often used for this.

MH4U Dex/QuestDetailsTableViewController.m
```diff
@@ -0,0 +1,189 @@
...
+    
+    NSSortDescriptor *descriptor = [[NSSortDescriptor alloc] initWithKey:MonsterAttributes.sort_name ascending:YES];
+    self.questMonsters = [self.quest.monster sortedArrayUsingDescriptors:@[descriptor]];
+}
+
+#pragma mark - Table view data source
+
+- (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView
+{
+    return self.questMonsters.count ? 2 : 1;
```
brockboland, line 72: Could you include a comment explaining what these numbers mean?

jmgoldb, line 72: Sure

MH4U Dex/QuestDetailsTableViewController.m
```diff
@@ -0,0 +1,189 @@
...
+            cell.textLabel.text = @"Large Monster";
+            cell.detailTextLabel.text = monster.name;
+            break;
+    }
+    return cell;
+}
+
+- (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    UIStoryboard *storyBoard = [UIStoryboard storyboardWithName:@"Main" bundle:nil];
```
brockboland: I'd also recommend a constant for the storyboard name

MH4U Dex/QuestDetailsTableViewController.m
```diff
@@ -0,0 +1,189 @@
...
+                } else {
+                    cell.detailTextLabel.text = number.stringValue;
+                }
+            } else {
+                cell.detailTextLabel.text = self.questDetailsDict[rowKey];
+            }
+            break;
+        case 1:
+            cell.userInteractionEnabled = YES;
+            monster = self.questMonsters[indexPath.row];
```
brockboland: You could be using your `monsterForIndexPath:` helper here

MH4U Dex/QuestListTableViewController.m
```diff
@@ -32,6 +34,17 @@ - (void)viewDidLoad
     self.fetchedResultsController.delegate = self;
 }
 
+- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
+{
+    QuestContainerViewController *questVC = (QuestContainerViewController *)segue.destinationViewController;
+    if ([segue.identifier isEqualToString:@"showQuestContainerView"]) {
```
brockboland: Same thing here about constants for segue identifier

MH4U Dex/QuestContainerViewController.m
jmgoldb: You mentioned I should put the constants in a struct, is there a particular reason for that?

brockboland: Just so they're grouped. You might access this one as `SegueIdentifiers.EmbeddedQuestDetails` and the other as `SegueIdentifiers.ShowQuestContainer`. I think that most of our projects use this strategy, and put the struct in the constants file.

jmgoldb: I'll take a look at how nextTier does it

brockboland:
A few suggestions.
1. In QuestDetailsTableViewController, I'd use an enum for the sections, just like you did in QuestContainerViewController. In this case, the count doesn't need to be part of the enum, since the number of sections is dependent on the contents of the `questMonsters` property. But, having a bunch of 0's and 1's in the various table view datasource/delegate methods isn't very clear.
2. You asked about the nested switch statements: yeah, that method is kind of a beast now. I would split that out into two helper methods that create and configure the two types of cells, and then call one of those depending on the section that the cell will go in.
3. The cell config for the first section is kinda funky. At the very least, I'd add comments explaining that (for example) the caravan/hall and other values are bool numbers, and so on.
4. The `QuestDetailEntries` and constant strings seem pretty verbose, but I'm not sure there's a much better way to handle it. I think I would combine them into an array, like this:

``` objc
NSArray *questDetailLabels = @[
                               @"Caravan/Hall",
                               @"Stars",
                               @"Objective",
                               @"Sub Objective",
                               @"Quest Type",
                               @"Key Quest",
                               @"Urgent Quest",
                               @"Unstable Quest",
                               @"Quest Fee",
                               @"Zenny Reward",
                               @"HRP Reward",
                               ];                           
```

Then, when building the cell:

``` objc
NSString *rowKey = questDetailLabels[indexPath.row];
```

There would still be some funky handling for the cases that use a boolean number, but it might help clean things up a bit.

jmgoldb:
Alrighty, I think that should take care of everything, but I feel weird calling on the indices of questDetailLabels directly.  Can you give me your opinion?

brockboland:
Looks a lot better. The `questDetailLabels` indices…yeah, I'm not sure how to handle that best. Maybe it would make sense to keep that big `switch` statement in `detailCellForIndexPath:` after all, and lose `questDetailsDict` completely.

jmgoldb:
I dunno, I think its far more readable without the giant switch statement.  And if don't use questDetailsDict, then I have to make things even more sprawling by manually assigning each property of the quest model object.

brockboland:
True. I'm fine with this as it is, but keep it at the back of your mind in case you come up with a better option.

Good to pull

jmgoldb:
Will do.  Thanks

jmgoldb:
@chillpop @seanwolter @designatednerd Can I get another merge, please?

seanwolter:
:suspect:

seanwolter:
That's the "suspect" face. 
I guess if Brock sez it's good to pull...

jmgoldb:
Isn't that the doomguy face?

seanwolter:
@jmgoldb They're all here for you, my friend: http://www.emoji-cheat-sheet.com

scottferg:
:neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard::neckbeard:

